PR #7, opened 2024-12-27
Fix the bug in QFT template

liwt31:


refraction-ray:
thanks for the contribution!

PR #8, opened 2025-01-16
Matrix square root for a not-perfectly PSD matrix

abhinavd:
This method zeros out any negative eigenvalues before taking the matrix square root. This ensures that the matrix square root of a known-to-be PSD matrix continues to have real eigenvalues. This is useful sometimes in cases where one wants something like $\sqrt{A^\dagger A}$ but $A^\dagger A$ is imperfectly computed and has numerically small negative eigenvalues.

### DIFF
--- a/tensorcircuit/backends/abstract_backend.py
+++ b/tensorcircuit/backends/abstract_backend.py
@@ -60,6 +60,21 @@ class ExtendedBackend:
         e = self.sqrt(e)
         return v @ self.diagflat(e) @ self.adjoint(v)
 
+    def sqrtmhpos(self: Any, a: Tensor) -> Tensor:
+        """
+        Return the sqrtm of a known-to-be PSD Hermitian matrix ``a``.
+
+        :param a: tensor in matrix form
+        :type a: Tensor
+        :return: sqrtm of ``a`` after setting the negative eigenvalues (if they exist) to zero
+        :rtype: Tensor
+        """
+        # maybe friendly for AD and also cosidering that several backend has no support for native sqrtm
+        e, v = self.eigh(a)
+        e = self.relu(e)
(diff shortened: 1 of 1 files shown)


tensorcircuit/backends/abstract_backend.py
```diff
@@ -60,6 +60,21 @@ def sqrtmh(self: Any, a: Tensor) -> Tensor:
...
+        :param a: tensor in matrix form
+        :type a: Tensor
+        :return: sqrtm of ``a`` after setting the negative eigenvalues (if they exist) to zero
+        :rtype: Tensor
+        """
+        # maybe friendly for AD and also cosidering that several backend has no support for native sqrtm
+        e, v = self.eigh(a)
+        e = self.relu(e)
+        e = self.sqrt(e)
+        return v @ self.diagflat(e) @ self.adjoint(v)
```
refraction-ray, line 76: what about directly including ``relu`` line in ``sqrtmh`` function implementation as non-PSD matrix has no well defined ``sqrtm`` anyway?

abhinavd: That should also work, although hopefully no one relies on the current behaviour of `sqrtmh`. That's why I thought it might be safest to have a different function itself. Or maybe add an optional keyword in `sqrtmh` to enforce PSD-ness.

refraction-ray: @abhinavd , ah, yes, `sqrt` can return complex values if the input is of negative value and in complex dtype. Then I think the better way is to add an optional keyword in `sqrtmh`, default as `psd=False`.

refraction-ray:
Thanks for the contribution, fixed now by e105763d92328c1ac1004fa30a79388c007bc6e3

PR #18, opened 2025-07-04
feat(templates): Implement unified lattice module and tests

Stellogic:
Hi @yutuer21,

Following your guidance and the OSPP project requirements ("Object-Oriented Lattice Generation"), I've prepared this initial pull request to add the unified lattice module. I would be grateful for your review and feedback.

**Main Changes in this PR:**

* **`AbstractLattice`:** An abstract base class is proposed to define the common API.
* **Concrete Implementations:** `CustomizeLattice` and `TILattice` (with `SquareLattice`, `HoneycombLattice`, `TriangularLattice` examples) are provided as initial implementations.
* **Core Functionality:** Basic methods like `get_site_info()`, `get_neighbors()`, and `.show()` visualization are included.
* **Testing:** A corresponding test suite (`tests/test_lattice.py`) has been added to cover the core functions and edge cases, achieving ~94% coverage on `lattice.py`.
* **Documentation:** `CHANGELOG.md` has been updated.

This is the first version based on my understanding of the requirements, and I'm sure there is still much room for improvement. Thank you very much for your valuable time and guidance. I look forward to your feedback!

### DIFF
--- a/examples/lattice_neighbor_benchmark.py
+++ b/examples/lattice_neighbor_benchmark.py
@@ -1,0 +1,98 @@
+"""
+An example script to benchmark neighbor-finding algorithms in CustomizeLattice.
+
+This script demonstrates the performance difference between the KDTree-based
+neighbor search and a baseline all-to-all distance matrix method.
+As shown by the results, the KDTree approach offers a significant speedup,
+especially when calculating for a large number of neighbor shells (large max_k).
+
+To run this script from the project's root directory:
+  python examples/templates/lattice_neighbor_benchmark.py
+"""
+
+import timeit
+from typing import Any, Dict, List
+
+
+def run_benchmark() -> None:
+    """
+    Executes the benchmark test and prints the results in a formatted table.
+    """
+    # --- Benchmark Parameters ---
+    # A list of lattice sizes (N = number of sites) to test
+    site_counts: List[int] = [10, 50, 100, 200, 500, 1000, 1500, 2000]
+
+    # Use a large k to better showcase the performance of KDTree in
+    # finding multiple neighbor shells, as suggested by the maintainer.
+    max_k: int = 2000
+
+    # Reduce the number of runs to keep the total benchmark time reasonable,
+    # especially with a large max_k.
+    number_of_runs: int = 3
+    # --------------------------
+
+    results: List[Dict[str, Any]] = []
+
+    print("=" * 75)
+    print("Starting neighbor finding benchmark for CustomizeLattice...")
+    print(f"Parameters: max_k={max_k}, number_of_runs={number_of_runs}")
+    print("=" * 75)
+    print(
+        f"{'Sites (N)':>10} | {'KDTree Time (s)':>18} | {'Baseline Time (s)':>20} | {'Speedup':>10}"
+    )
+    print("-" * 75)
+
+    for n_sites in site_counts:
+        # Prepare the setup code for timeit.
+        # This code generates a random lattice and is executed before timing begins.
+        # We use a fixed seed to ensure the coordinates are the same for both tests.
+        setup_code = f"""
+import numpy as np
+from tensorcircuit.templates.lattice import CustomizeLattice
+
+np.random.seed(42)
+coords = np.random.rand({n_sites}, 2)
+ids = list(range({n_sites}))
+lat = CustomizeLattice(dimensionality=2, identifiers=ids, coordinates=coords)
+"""
+        # Define the Python statements to be timed.
+        stmt_kdtree = f"lat._build_neighbors(max_k={max_k})"
+        stmt_baseline = f"lat._build_neighbors_by_distance_matrix(max_k={max_k})"
+
+        try:
+            # Execute the timing. timeit returns the total time for all runs.
+            time_kdtree = (
+                timeit.timeit(stmt=stmt_kdtree, setup=setup_code, number=number_of_runs)
+                / number_of_runs
+            )
+            time_baseline = (
+                timeit.timeit(
+                    stmt=stmt_baseline, setup=setup_code, number=number_of_runs
+                )
+                / number_of_runs
+            )
+
+            # Calculate and store results, handling potential division by zero.
+            speedup = time_baseline / time_kdtree if time_kdtree > 0 else float("inf")
+            results.append(
+                {
+                    "n_sites": n_sites,
+                    "time_kdtree": time_kdtree,
+                    "time_baseline": time_baseline,
+                    "speedup": speedup,
+                }
+            )
+            print(
+                f"{n_sites:>10} | {time_kdtree:>18.6f} | {time_baseline:>20.6f} | {speedup:>9.2f}x"
+            )
+
+        except Exception as e:
+            print(f"An error occurred at N={n_sites}: {e}")
+            break
+
+    print("-" * 75)
+    print("Benchmark complete.")
+
+
+if __name__ == "__main__":
+    run_benchmark()
--- a/tensorcircuit/templates/__init__.py
+++ b/tensorcircuit/templates/__init__.py
@@ -5,5 +5,6 @@ from . import dataset
 from . import graphs
 from . import measurements
 from . import conversions
+from . import lattice
 
 costfunctions = measurements
--- a/tensorcircuit/templates/lattice.py
+++ b/tensorcircuit/templates/lattice.py
@@ -474,12 +474,14 @@ class AbstractLattice(abc.ABC):
         :return: A sorted list of squared distances representing the shells.
         :rtype: List[float]
         """
+        ZERO_THRESHOLD_SQ = 1e-12
+
         all_distances_sq = np.asarray(all_distances_sq)
         # Now, the .size call below is guaranteed to be safe.
         if all_distances_sq.size == 0:
             return []
 
-        sorted_dist = np.sort(all_distances_sq[all_distances_sq > 1e-12])
+        sorted_dist = np.sort(all_distances_sq[all_distances_sq > ZERO_THRESHOLD_SQ])
 
         if sorted_dist.size == 0:
             return []
@@ -684,18 +686,14 @@ class TILattice(AbstractLattice):
     def _build_neighbors(self, max_k: int = 2, **kwargs: Any) -> None:
         """Calculates neighbor relationships for the periodic lattice.
 
-        This optimized method leverages the translational symmetry of the lattice.
-        It first calculates neighbor relationships for only the reference sites
-        in the first unit cell (uc_coord=(0,...)). These relationships, defined
-        by unit cell displacements and basis site changes, serve as a template.
-        This template is then applied (translated) to all other unit cells in
-        the lattice to determine their neighbors efficiently. This avoids the
-        costly O(N^2) full distance matrix calculation for *finding neighbors*.
-        Note: For non-fully periodic systems, this falls back to a robust
-        distance-based method applying the Minimum Image Convention.
+        This method calculates neighbor relationships by computing the full N x N
+        distance matrix. It robustly handles all boundary conditions (fully
+        periodic, open, or mixed) by applying the Minimum Image Convention
+        (MIC) only to the periodic dimensions.
 
-        Finally, the full N x N distance matrix (respecting PBC) is always
-        computed and cached at the end of this process.
+        From this distance matrix, it identifies unique neighbor shells up to
+        the specified `max_k` and populates the neighbor maps. The computed
+        distance matrix is then cached for future use.
 
         :param max_k: The maximum number of neighbor shells to
             calculate. Defaults to 2.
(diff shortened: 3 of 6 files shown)


tensorcircuit/templates/lattice.py
```diff
@@ -0,0 +1,886 @@
...
+    def __len__(self) -> int:
+        """Returns the total number of sites, enabling `len(lattice)`."""
+        return self.num_sites
+
+    # --- Public API for Accessing Lattice Information ---
+
+    def get_coordinates(self, index: SiteIndex) -> Coordinates:
+        """Gets the spatial coordinates of a site by its integer index.
+
+        Args:
```
refraction-ray, line 78: for the format of the docstring, please refer to other files in this repo, e.g., https://github.com/tensorcircuit/tensorcircuit-ng/blob/master/tensorcircuit/quantum.py#L1327-L1359

tensorcircuit/templates/lattice.py
```diff
@@ -0,0 +1,886 @@
...
+    Tuple,
+    Union,
+    TYPE_CHECKING,
+)
+
+import matplotlib.pyplot as plt
+
+import numpy as np
+
+if TYPE_CHECKING:
```
refraction-ray, line 22: what mypy error is this used for?

tensorcircuit/templates/lattice.py
```diff
@@ -0,0 +1,886 @@
...
+                `matplotlib.pyplot.scatter` function for customizing site appearance.
+        """
+        # creat "fig_created_internally" as flag
+        fig_created_internally = False
+
+        if self.num_sites == 0:
+            print("Lattice is empty, nothing to show.")
+            return
+        if self.dimensionality not in [1, 2, 3]:
+            print(
```
refraction-ray, line 299: better directly use ``warning`` or ``logger.warning`` instead of print

refraction-ray:
some more TILattice, 
1D: chain, chain with AB sublattice (ABABAB...), 
2D: rectangular,  checkboard (i.e. square lattice with AB sublattice), Lieb lattice, Kagome lattice 
3D: cubic lattice

refraction-ray:
please also modify __init__.py in templates module accordingly

Stellogic:
Hi @refraction-ray,

Thank you again for your invaluable guidance and detailed review. I have now pushed the updates, incorporating all of your feedback.

Here is a summary of the main changes:

*   **Logging & On-Demand Computation**: Replaced all `print` calls with `logging` and implemented on-demand neighbor calculations.
*   **New `TILattice` Classes**: Added and tested all the suggested lattice implementations (`Chain`, `Kagome`, `Lieb`, etc.).
*   **Code Style & Typing**:
    *   Removed type hints from test function arguments as requested.
    *   In response to your question about the `mypy` error, I have added a detailed comment to the source code explaining the purpose of the `if TYPE_CHECKING:` block for handling the optional `matplotlib` dependency.
*   **Docstring Format Correction**: I have carefully revised all docstrings in `lattice.py` to match the project's required  style.

tensorcircuit/templates/__init__.py
```diff
@@ -7,3 +7,16 @@
 from . import conversions
 
 costfunctions = measurements
+from .lattice import AbstractLattice
```
refraction-ray, line 10: just `from . import lattice` as above

tensorcircuit/templates/lattice.py
```diff
@@ -0,0 +1,1198 @@
...
+            comparisons. Defaults to 1e-6.
+        :type tol: float, optional
+        """
+        self._neighbor_maps = {k: {} for k in range(1, max_k + 1)}
+        if self.num_sites < 2:
+            return
+
+        potential_neighbors: Dict[int, List[Tuple[float, int]]] = {}
+        relative_cell_shifts = (
+            np.array([p for p in np.ndindex(*((3,) * self.dimensionality))]) - 1
```
refraction-ray, line 641: how can the radius 3 here is sufficient for large `max_k`? the search radius should be propotional to `max_k`?

tensorcircuit/templates/lattice.py
```diff
@@ -0,0 +1,1198 @@
+# -*- coding: utf-8 -*-
+"""
+The lattice module for defining and manipulating lattice geometries.
+"""
+import logging
+
```
refraction-ray, line 6: no empty lines here

tensorcircuit/templates/lattice.py
```diff
@@ -0,0 +1,1198 @@
...
+            np.array([p for p in np.ndindex(*((3,) * self.dimensionality))]) - 1
+        )
+
+        for i in range(self.num_sites):
+            # Use a temporary set for each site to automatically handle duplicates
+            neighbors_set: set[Tuple[float, int]] = set()
+
+            ident_i = self._identifiers[i]
+            coord_i = self._coordinates[i]
+            assert isinstance(ident_i, tuple)
```
refraction-ray, line 650: no need to assert on variables of the class, such attrs binding should be checked when initialization the class instance if necessary.

tests/test_lattice.py
```diff
@@ -0,0 +1,883 @@
...
+        lattice_4d = CustomizeLattice(
+            dimensionality=4, identifiers=[0], coordinates=[[0, 0, 0, 0]]
+        )
+
+        # Act
+        lattice_4d.show()
+
+        # Assert: Check that the appropriate warning was printed to stdout.
+        assert "show() is not implemented for 4D lattices." in caplog.text
+
```
refraction-ray, line 850: I suggest to add one or few tests on relative large lattice, say $6 \times 8$, and to check whether neighbor searching is correctly handled for large `max_k`, say 4, 6, 8 and more, what if the given `max_k` is larger than the possible `max_k`?

refraction-ray:
Thanks for the nice contribution. In general, the code looks very good.

The addition should be a more robust and error-free neighbor searching implementation with careful consideration on PBC and large `max_k`

current failed case for neighbor search:
```python
sq = lattice.SquareLattice(size=(4,4), pbc=True)
len(sq.get_neighbor_pairs(k=3))
# 0
```

tensorcircuit/templates/lattice.py
```diff
@@ -0,0 +1,1403 @@
...
+            identifier next to each site. Defaults to False.
+        :type show_identifiers: bool, optional
+        :param show_bonds_k: Specifies which order of
+            neighbor bonds to draw (e.g., 1 for NN, 2 for NNN). If None,
+            no bonds are drawn. If the specified neighbors have not been
+            calculated, a warning is printed. Defaults to 1.
+        :type show_bonds_k: Optional[int], optional
+        :param ax: An existing Matplotlib Axes object to plot on.
+            If None, a new Figure and Axes are created automatically. Defaults to None.
+        :type ax: Optional["matplotlib.axes.Axes"], optional
```
refraction-ray, line 302: no doc for `bond_kwards` arg

Stellogic: Got it, thanks for the review! I'll add the docstring now.

Stellogic:
Hi @refraction-ray 
Thanks for the great suggestions! I've now pushed all the updates.
I've implemented the dynamic modification APIs (from_lattice, add_sites, remove_sites) for CustomizeLattice.
I've fixed the default show() behavior to avoid warnings.

I ran a performance benchmark for the neighbor finding methods. The data shows that the KDTree-based method is likely consistently faster. Based on this, I've kept the implementation simple, using KDTree by default without needing a user-facing switch. The benchmark script is included in templates/ .benchmarks/.

New tests for all new features have been added.

tensorcircuit/templates/lattice.py
```diff
@@ -0,0 +1,1539 @@
...
+                )
+
+                # Calculate squared distances for all images
+                image_d_sq = np.sum(image_displacements**2, axis=2)
+
+                # The true distance is the minimum over all images
+                dist_matrix_sq[i, :] = np.min(image_d_sq, axis=1)
+
+            # Part C: Identify neighbor shells from the correctly calculated distance matrix.
+            # This logic is copied from the working `_find_neighbors_by_distance` method.
```
refraction-ray, line 707: `self._distance_matrix` assignment for TILattice mixed boundary condition case

tensorcircuit/templates/lattice.py
```diff
@@ -0,0 +1,1539 @@
...
+                "For better performance, please install it using 'pip install scipy'."
+            )
+            # Fallback to the original, robust-but-slower method
+            return self._build_neighbors_by_distance_matrix(max_k=max_k, tol=tol)
+
+        all_coords = np.array(self._coordinates)
+
+        # 1. Use pdist for memory-efficient calculation of pairwise distances
+        #    to robustly identify the distance shells.
+        all_distances_sq = pdist(all_coords, metric="sqeuclidean")
```
refraction-ray, line 1370: `self._distance_matrix` assignment for CustomizeLattice KDTree engine case

refraction-ray:
1. I have added comments for positions to add `self._distance_matrix` assignment for different neighbor building functions, still, the `TILattice` fully PBC case requires several further lines for generating the matrix when build neighbors as the all-to-all distance is not explicitly generated in that case.
2. scipy is always assumed installed as required by `setup.py`, so you can safely delete the check and fallback logic for whether scipy is installed
3. you can move the benchmark script as one example scripts in `/examples` folders

after addressing these points, I believe this PR is perfect to be merged. later you can open new PRs for more examples, tutorials, docs, or feature enhancements.

tensorcircuit/templates/.benchmarks/benchmark_neighbors.py
```diff
@@ -0,0 +1,86 @@
...
+logging.getLogger("tensorcircuit.templates.lattice").setLevel(logging.WARNING)
+
+
+def run_benchmark() -> None:
+    """
+    Executes the benchmark test and prints the results in a formatted table.
+    """
+    site_counts: List[int] = [10, 50, 100, 200, 500, 1000, 1500, 2000]
+    max_k: int = 1
+    number_of_runs: int = 5
```
refraction-ray, line 27: better reduce to 3 for the sake of the total running time of the script

examples/lattice_neighbor_benchmark.py
```diff
@@ -0,0 +1,102 @@
...
+
+To run this script from the project's root directory:
+  python examples/templates/lattice_neighbor_benchmark.py
+"""
+import logging
+import timeit
+from typing import Any, Dict, List
+
+# Suppress verbose logging from the lattice module during the benchmark
+logging.getLogger("tensorcircuit.templates.lattice").setLevel(logging.WARNING)
```
refraction-ray, line 18: the info level log is ignored by default, I guess removing this line is fine?

examples/lattice_neighbor_benchmark.py
```diff
@@ -0,0 +1,102 @@
+# examples/templates/lattice_neighbor_benchmark.py
```
refraction-ray, line 1: remove this line

tensorcircuit/templates/lattice.py
```diff
@@ -0,0 +1,1555 @@
...
+                    if is_in_bounds:
+                        target_ident = tuple(target_uc) + (target_basis,)
+                        # The identifier must exist in the map
+                        if target_ident in self._ident_to_idx:
+                            neighbor_idx = self._ident_to_idx[target_ident]
+                            neighbors_of_i_for_k.append(neighbor_idx)
+
+                if neighbors_of_i_for_k:
+                    self._neighbor_maps[k][i] = sorted(neighbors_of_i_for_k)
+        logger.info("Caching the full distance matrix with Minimum Image Convention...")
```
refraction-ray, line 831: this `distance_matrix` should be obtained more cleverly, otherwise it will ruin the advantage for an independent neighbor search logic for fully PBC cases, as you again recompute the full distance_matrix from the scratch

tensorcircuit/templates/lattice.py
refraction-ray: again `distance_matrix` should be assembled from `all_distances_sq`, otherwise, the all-to-all distance is computed twice?

Stellogic:
@refraction-ray

Thanks for the review! I've addressed all the feedback and squashed the commits.

tensorcircuit/templates/lattice.py
refraction-ray: distance matrix should also be assigned here, so that after neighbor building, you dont need to recompute `pdist` by calling `_compute_distance_matrix`. The logic, lattice class without neighbor building, distance matrix can be accessed by `_compute_distance_matrix`, lattice class after neighbor building can directly return the matrix without recompute via `pdist` in `_compute_distance_matrix`

tests/test_lattice.py
```diff
@@ -0,0 +1,1276 @@
...
+    def test_modification_clears_distance_matrix_cache(self, initial_lattice):
+        """
+        Tests that add_sites and remove_sites correctly invalidate the
+        cached distance matrix.
+        """
+        # Arrange 1: Compute neighbors and distance matrix
+        lat = initial_lattice
+        lat._build_neighbors(max_k=1)
+        _ = lat.distance_matrix
```
refraction-ray, line 1185: a meaningfull test on the matrix instead of a plain assignment?

refraction-ray:
The only remaining part I believe is the improvement on the logic for `distance_matrix` caching when building neighbors. For `CustomizeLattice` case, a simply assignment during neighbor searching could achieve this. For fully PBC case of `TILattice`, the assignment is still missing, hopefully, we can reconstruct the distance matrix based on all the infos from neighbor searching without the direct computation of all-to-all distance matrix. The current independent implementation design for fully PBC case is trying to avoid the full matrix computation after all.

tensorcircuit/templates/lattice.py
```diff
@@ -0,0 +1,1526 @@
...
+        :type max_k: int
+        :param tol: The numerical tolerance to consider two distances equal.
+        :type tol: float
+        :return: A sorted list of squared distances representing the shells.
+        :rtype: List[float]
+        """
+        if not isinstance(all_distances_sq, np.ndarray):
+            all_distances_sq = np.array(all_distances_sq)
+
+        sorted_dist_sq = np.sort(all_distances_sq[all_distances_sq > 1e-12])
```
refraction-ray, line 480: why not tol**2 but an explicit value, no need to change the code for this function?

tensorcircuit/templates/lattice.py
```diff
@@ -0,0 +1,1526 @@
...
+            )
+            for source_basis_idx, target_site_idx in zip(*match_indices):
+                source_ident = self._identifiers[source_basis_idx]
+                target_ident = self._identifiers[target_site_idx]
+                uc_disp = np.array(target_ident[:-1]) - np.array(source_ident[:-1])
+                target_basis = target_ident[-1]
+                neighbor_templates[source_basis_idx][k].append((uc_disp, target_basis))
+
+        # 4. Apply templates and SIMULTANEOUSLY reconstruct the distance matrix.
+        self._neighbor_maps = {k: {} for k in range(1, max_k + 1)}
```
refraction-ray, line 770: why changing this? the original `self._neighbor_maps = {k: {} for k in range(1, len(dist_shells_sq) + 1)}` is better? which could also free L803-L809, or am I wrong?

tensorcircuit/templates/lattice.py
```diff
@@ -0,0 +1,1526 @@
...
+        if not dist_shells_sq:
+            logger.info("No distinct neighbor shells found.")
+            return
+
+        tree = KDTree(all_coords)
+        self._neighbor_maps = {}
+
+        found_neighbors_so_far: List[set[int]] = [set() for _ in range(self.num_sites)]
+
+        for k_idx, target_d_sq in enumerate(dist_shells_sq):
```
refraction-ray, line 1347: why the neigbor building logic here is also modified? The original implementation is more neat? what is the gain here?

tests/test_lattice.py
```diff
@@ -0,0 +1,1540 @@
...
+        # Assert
+        # We compare directly against a pre-sorted list, not a set.
+        # This will fail if the implementation returns [3, 0] instead of [0, 3].
+        assert neighbors == [
+            0,
+            3,
+        ], "The neighbor list should be sorted in ascending order."
+
+
+class TestDistanceMatrixStrength:
```
refraction-ray, line 1390: the name convention with Strength is weird, why not only TestDistanceMatrix or something alike?

Stellogic:
@refraction-ray 
My apologies. My process before updating a PR has been to let an AI review it first for any potential issues. I would then follow its suggestions to modify things like variable names, logic, and docstrings. Going forward, I will be more careful in reviewing the AI's suggestions to ensure they are reasonable. Again, I am very sorry."、

Stellogic:
@refraction-ray
My apologies. My process before updating a PR has been to let an AI review it first for any potential issues. I would then follow its suggestions to modify things like variable names, logic, and docstrings. Going forward, I will be more careful in reviewing the AI's suggestions to ensure they are reasonable. Again, I am very sorry.

refraction-ray:
@Stellogic , No need to apologize. It is super okay to use AI, actually I strongly recommend you guys to intensively use AI for coding and development tasks. Just pay attention to the points I mention, as a large bulk of unnecessary changes will make the code review more difficult. AI can write the code for the most parts and everyone use it. It is a good thing, but you should be the first responsible reviewer for what AI generates and changes :)

Stellogic:
Hi @refraction-ray,

Following your guidance, I have force-pushed an update. I've reset the branch back to commit 9ef578d and built a single, clean commit on top of it containing only the necessary fixes to address the test failures.

This new commit includes three targeted fixes:

Fixed Distance Matrix Caching for TILattice: In the fully periodic case, _build_neighbors now reconstructs and caches the full distance matrix by leveraging the existing ref_dist_matrix_sq and translational symmetry. This resolves all failures in TestDistanceMatrix.

Fixed Neighbor Finding for Duplicate Coordinates in 
CustomizeLattice: The logic in _build_neighbors has been refined. It now correctly handles co-located sites by ensuring all points at distance ~0 are excluded from the initial k=1 neighbor shell, which fixes the test_lattice_with_duplicate_coordinates failure.

Improved Numerical Stability in _identify_distance_shells: The logic for filtering zero-distance points has been decoupled from the tolerance-based shell identification. It now robustly filters out self-loops using a fixed small threshold (1e-12), resolving the test_neighbor_shells_with_tiny_separation failure.

tensorcircuit/templates/lattice.py
```diff
@@ -747,6 +744,7 @@ def _build_neighbors(self, max_k: int = 2, **kwargs: Any) -> None:
                     current_k_map[i].sort()
 
                 self._neighbor_maps[k] = current_k_map
+                self._distance_matrix = dist_matrix
```
refraction-ray, line 747: wrong tab? why not moving this line directly below L717?

tensorcircuit/templates/lattice.py
```diff
@@ -479,18 +479,15 @@ def _identify_distance_shells(
         if all_distances_sq.size == 0:
             return []
 
-        # Sort unique distances and filter out zero-distance (self-loops)
-        unique_sorted_dist = sorted(
-            [d for d in np.unique(all_distances_sq) if d > tol**2]
-        )
+        sorted_dist = np.sort(all_distances_sq[all_distances_sq > 1e-12])
```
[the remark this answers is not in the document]
Stellogic: I've looked into this, and it seems that a direct replacement with tol**2 causes the test_neighbor_shells_with_tiny_separation test to fail.
 The root cause seemms that tol is treated as a linear distance tolerance, but using tol**2 in the squared-distance comparison logic makes the shell-merging behavior overly sensitive and breaks the test's intent.

refraction-ray:
the CI error seems to be irrelevant for this PR. However, I am still confused on the success of `test_neighbor_shells_with_tiny_separation` function

tensorcircuit/templates/lattice.py
```diff
@@ -0,0 +1,1603 @@
...
+        for i in range(self.num_sites):
+            ident_i = cast(Tuple[Any, ...], self._identifiers[i])
+            uc_i = np.array(ident_i[:-1])
+            basis_i = ident_i[-1]
+
+            uc_disp = -uc_i
+            relative_ucs_list = [
+                cast(Tuple[Any, ...], ident)[:-1] for ident in self._identifiers
+            ]
+            relative_ucs_arr = np.array(relative_ucs_list)
```
refraction-ray, line 864: should this be outside the for as it doesn't depend on i?

tensorcircuit/templates/lattice.py
```diff
@@ -0,0 +1,1603 @@
...
+                source_ident = self._identifiers[source_basis_idx]  # (0,..,0,basis)
+                target_ident = self._identifiers[target_site_idx]
+
+                # Rule: (uc_displacement, target_basis_index)
+                uc_disp = np.array(target_ident[:-1]) - np.array(source_ident[:-1])
+                target_basis = target_ident[-1]
+                neighbor_templates[source_basis_idx][k].append((uc_disp, target_basis))
+
+        # 4. Apply the templates to all sites in the lattice
+        self._neighbor_maps = {k: {} for k in range(1, len(dist_shells_sq) + 1)}
```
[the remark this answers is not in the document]
refraction-ray: currently both pbc and obc in TILattice automatically build the whole neighbor maps with invalid `max_k` or `precompute_neighbors`

refraction-ray:
image
my suggestion is to not use the special implementation of fully PBC case, as it is much slower and greatly complicate the implementation

refraction-ray:
the mixed boundary condition universal implementation is efficient enough, just deleting the special tailored implementation branch for fully PBC case, and everything would be nicer and easier

refraction-ray:
@all-contributors please add @Stellogic for code, examples, tests

refraction-ray:
Apart from some more tutorials and examples, the next key milestone could be a machine learning backend agnostic implementation for lattice module, compared to the numpy backend implementation now. A backend agnostic implementation using tensorcircuit.backend API would enable automatic differentiation, vectorized map etc. for this lattice module. A typical example is structure optimization, where the energy computed from the lattice is minimized via gradient descent with tunable parameters the lattice structure parameters.

PR #22, opened 2025-07-24
Feature/hamiltonian generation

Stellogic:
This pull request implements **Task 1** of the follow-up plan, introducing utility functions to generate common spin Hamiltonians from `Lattice` objects.

### Key Contributions:

- **`generate_heisenberg_hamiltonian`**: Creates a sparse Heisenberg Hamiltonian based on the nearest-neighbor pairs provided by the lattice.
- **`generate_rydberg_hamiltonian`**: Creates a sparse Rydberg Hamiltonian, utilizing the lattice's distance matrix to calculate long-range interactions.
- **Comprehensive Tests**: Added a new test file (`tests/templates/test_hamiltonians.py`) with unit tests that verify:
    - Scientific correctness against manually calculated small systems.
    - Robustness against edge cases (empty lattice, single site, zero distance).
    - Fundamental physical properties (e.g., Hermiticity).

### DIFF
--- a/tensorcircuit/templates/hamiltonians.py
+++ b/tensorcircuit/templates/hamiltonians.py
@@ -1,7 +1,6 @@
-import typing
 from typing import Any, List, Tuple, Union
 import numpy as np
-from tensorcircuit.cons import dtypestr, backend
+from ..cons import dtypestr, backend
 from ..quantum import PauliStringSum2COO
 from .lattice import AbstractLattice
 
@@ -53,8 +52,8 @@ def heisenberg_hamiltonian(
 
     pauli_map = {"X": 1, "Y": 2, "Z": 3}
 
-    ls: typing.List[typing.List[int]] = []
-    weights: typing.List[float] = []
+    ls: List[List[int]] = []
+    weights: List[float] = []
 
     pauli_terms = ["X", "Y", "Z"]
     for i, j in neighbor_pairs:
@@ -102,8 +101,8 @@ def rydberg_hamiltonian(
         raise ValueError("Cannot generate a Hamiltonian for a lattice with zero sites.")
 
     pauli_map = {"X": 1, "Y": 2, "Z": 3}
-    ls: typing.List[typing.List[int]] = []
-    weights: typing.List[float] = []
+    ls: List[List[int]] = []
+    weights: List[float] = []
 
     for i in range(num_sites):
         x_string = [0] * num_sites
--- a/tensorcircuit/templates/hamiltonians.py
+++ b/tensorcircuit/templates/hamiltonians.py
@@ -1,8 +1,9 @@
-import tensorcircuit as tc
+import typing
+from typing import cast
 from scipy.sparse import coo_matrix
 import numpy as np
+import tensorcircuit as tc
 
-import typing
 
 from .lattice import AbstractLattice
 
@@ -57,13 +58,9 @@ def generate_heisenberg_hamiltonian(
         ls.append(zz_string)
         weights.append(j_coupling)
 
-    hamiltonian_matrix = tc.quantum.PauliStringSum2COO(
-        ls, weight=weights, numpy=True
-    )
+    hamiltonian_matrix = tc.quantum.PauliStringSum2COO(ls, weight=weights, numpy=True)
 
-    return hamiltonian_matrix
-
-
+    return cast(coo_matrix ,hamiltonian_matrix)
 
 
 def generate_rydberg_hamiltonian(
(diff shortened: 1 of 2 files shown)


tensorcircuit/templates/hamiltonians.py
```diff
@@ -0,0 +1,139 @@
+import typing
+from typing import cast
+from scipy.sparse import coo_matrix
+import numpy as np
+import tensorcircuit as tc
+from .lattice import AbstractLattice
+
+
+def generate_heisenberg_hamiltonian(
```
refraction-ray, line 9: `def heisenberg_hamiltonian`

tensorcircuit/templates/hamiltonians.py
```diff
@@ -0,0 +1,139 @@
...
+    lattice: AbstractLattice, j_coupling: float = 1.0
```
refraction-ray, line 10: j_coupling is better be a vector whose dimension can be 1, 3, representing for 1: $J_x=J_y=Jz$; 3 different $J_{x, y, z}$

tensorcircuit/templates/hamiltonians.py
```diff
@@ -0,0 +1,139 @@
...
+            interaction_strength = c6 / (distance**6)
+            coefficient = interaction_strength / 4.0
+
+            zz_string = [0] * num_sites
+            zz_string[i] = pauli_map["Z"]
+            zz_string[j] = pauli_map["Z"]
+            ls.append(zz_string)
+            weights.append(coefficient)
+
+            z_coefficients[i] -= coefficient
```
refraction-ray, line 127: what is this for?

tests/test_hamiltonians.py
```diff
@@ -0,0 +1,132 @@
...
+        h_expected = j_coupling * (xx + yy + zz)
+
+        assert h_generated.shape == (4, 4)
+        assert np.allclose(h_generated.toarray(), h_expected)
+
+    def test_square_lattice_properties(self):
+        """
+        Test properties of a larger lattice (2x2 square) without full matrix comparison.
+        """
+        lattice = SquareLattice(size=(2, 2), pbc=True)  # 4 sites, 8 bonds with PBC
```
refraction-ray, line 67: 2*2 with PBC is subtle, as you can see, there will be two bonds between the same two lattice points

tensorcircuit/templates/hamiltonians.py
```diff
@@ -0,0 +1,139 @@
...
+from typing import cast
+from scipy.sparse import coo_matrix
+import numpy as np
+import tensorcircuit as tc
+from .lattice import AbstractLattice
+
+
+def generate_heisenberg_hamiltonian(
+    lattice: AbstractLattice, j_coupling: float = 1.0
+) -> coo_matrix:
```
refraction-ray, line 11: for simplicity, return type coo_matrix=Any is acceptable

tensorcircuit/templates/hamiltonians.py
```diff
@@ -84,12 +92,12 @@ def generate_rydberg_hamiltonian(
...
     :type c6: float
-    :return: The Hamiltonian represented as a SciPy COO sparse matrix.
-    :rtype: coo_matrix
+    :return: The Hamiltonian as a backend-agnostic sparse matrix.
+    :rtype: Any
     """
     num_sites = lattice.num_sites
     if num_sites == 0:
-        return coo_matrix((0, 0))
+        return _create_empty_sparse_matrix(shape=(1, 1))
```
refraction-ray, line 100: similarly, just raise ValueError for illegal number of sites

tensorcircuit/templates/hamiltonians.py
refraction-ray: ah, i see, my bad

Stellogic:
Hi @refraction-ray,
Thank you for the detailed and helpful review! I have updated the code according to your feedback.
Here is a summary of the changes:
Internal API Usage: I've updated the code to use backend from .cons and relative imports for internal modules, instead of calling the top-level tensorcircuit package.
Error Handling: I've added a ValueError check for num_sites = 0 in both Hamiltonian functions, as you suggested.
Rydberg Hamiltonian Sign: Regarding your question on the Z-term sign (why -v_ij/4?), I've double-checked the derivation and I thought the implementation is correct. The single-body Z coefficient comes from two sources:
The detuning term -delta * n_i gives a + (delta / 2) * Z_i contribution.
The interaction term V_ij * n_i * n_j expands to (V_ij / 4) * (1 - Z_i - Z_j + Z_i*Z_j), which contributes - (V_ij / 4) * Z_i to the single-body term.
I believe this addresses all your feedback. The PR should be ready for another look. Thanks again

tensorcircuit/templates/hamiltonians.py
```diff
@@ -0,0 +1,154 @@
+import typing
+from typing import Any, List, Tuple, Union
+import numpy as np
+from tensorcircuit.cons import dtypestr, backend
```
refraction-ray, line 4: should also be `from ..cons` ?

tensorcircuit/templates/hamiltonians.py
```diff
@@ -0,0 +1,154 @@
+import typing
```
refraction-ray, line 1: should merge with the following line? import all types you will use

Stellogic:
Thanks for the review, @refraction-ray! I've updated the PR according to your suggestions. It should be ready for another look now.

refraction-ray:
great! LGTM. will merge if CI finished successfully.

PR #24, opened 2025-07-29
feat(templates): Add greedy algorithm for gate layering

Stellogic:
Introduces the `get_nn_gate_layers` utility function. This function takes a lattice object and partitions its nearest-neighbor pairs into the minimum number of compatible layers for parallel two-qubit gate application.

This implementation uses a greedy edge-coloring algorithm to ensure that no two gates within the same layer act on the same qubit. The output is deterministic, with both layers and the edges within them being sorted.

This functionality is essential for efficiently scheduling gates in algorithms like Trotterized Hamiltonian evolution and directly addresses Task 2 of the lattice API follow-up plan.


Stellogic:
Hi @refraction-ray,

I've just opened this Pull Request for Task 2, which implements the `get_nn_gate_layers` function as we discussed.

This should be ready for your review when you have a moment. Please let me know if you have any suggestions or if any changes are needed.

Thanks again for your guidance!

refraction-ray:
some general suggestions:
1. just add the helper function in `tc.tempaltes.lattice`, no need for new independent test and moduel files
2. the function signature better be `get_compatible_layers(List[Tuple[int, int]])`. By this function, we can not only get compatible layers of NN bonds, we can also similarly get NNN bonds arangement etc.
3. with the lattice class and this helper function, you can implement a 2d circuit construction and vqe example in `/examples` similar to https://github.com/tensorcircuit/tensorcircuit-ng/blob/master/examples/vqe2d_gpu.py, where the Grid2D_entangling blocks are replaced by individual two qubit gates, say `c.rzz(i, j, theta=)` where i, j are neighbor pairs of given lattice arranged in the order compatible layers by `get_compatible_layers` helper function

examples/vqe2d_lattice.py
```diff
@@ -0,0 +1,81 @@
...
+import optax
+import tensorcircuit as tc
+from tensorcircuit.templates.lattice import SquareLattice, get_compatible_layers
+from tensorcircuit.templates.hamiltonians import heisenberg_hamiltonian
+
+# Use JAX for high-performance, especially on GPU.
+K = tc.set_backend("jax")
+tc.set_dtype("complex64")
+# On Windows, cotengra's multiprocessing can cause issues.
+tc.set_contractor("cotengra-8192-8192", parallel=False)
```
refraction-ray, line 11: parallel=False is subtle (very slow), maybe you can try other parallel options to see whether they work in windows? https://cotengra.readthedocs.io/en/latest/advanced.html#parallelization

tests/test_lattice.py
```diff
@@ -1,5 +1,6 @@
 from unittest.mock import patch
 import logging
+from typing import List, Set, Tuple
```
refraction-ray, line 3: we dont need to add type annotation for test functions, though it is okay to have them

examples/vqe2d_lattice.py
```diff
@@ -0,0 +1,81 @@
...
+        Defines the VQE ansatz and computes the energy expectation.
+        The ansatz consists of nlayers of RZZ, RXX, and RYY entangling layers.
+        """
+        c = tc.Circuit(n * m)
+        c = singlet_init(c)
+
+        for i in range(nlayers):
+            for layer in gate_layers:
+                for j, k in layer:
+                    c.rzz(int(j), int(k), theta=param[i, 0])
```
refraction-ray, line 43: we should find a better way to make each parameter different for each gate, a naive and also bruteforce way is to use a one-dimensional tensor for `param`, and for each use of `param`, we call `param[count]` then `count+=1`

tests/test_lattice.py
```diff
@@ -1664,3 +1667,113 @@ def test_distance_matrix_invariants_for_all_lattice_types(self, lattice):
 #             "The specialized PBC implementation is significantly slower "
 #             "than the general-purpose implementation."
 #         )
+
+
+class MockLattice(AbstractLattice):
```
refraction-ray, line 1672: what is this class for, I dont see it being used, maybe deleted?

refraction-ray:
overall, very nice. Leave some minor comments to be fixed.

PR #42, opened 2025-08-29
Qudit infras

WeiguoMa:
This PR contains extensive preliminary work for quditcircuit, including:
	1.	Removing and improving a large amount of hard-coded logic based on qubit systems with dimension 2, replacing it with a general dimension d;
	2.	Preserving the efficient sampling method for d = 2, while providing an if branch for cases where d != 2;
	3.	Updating the state labeling scheme from standard binary to a base-d representation using 0–9A–Z;
	4.	Adjusting the hard-coded parts in MPSCircuit and densitycircuit to ensure their functionality as qudit backends in the future, while currently forbidding the direct creation of qudits with d != 2 from mpscircuit and densitycircuit.


tensorcircuit/results/counts.py
```diff
@@ -105,44 +112,90 @@ def count2vec(count: ct, normalization: bool = True) -> Tensor:
...
+        if len(s) != n:
+            raise ValueError(
+                f"The length of all keys should be the same ({n}), received '{k}'."
+            )
+        for ch in s:
+            if ch not in _ALPHABET:
+                raise ValueError(
+                    f"Key '{k}' contains illegal character '{ch}' (only 0-9A-Z are allowed)."
+                )
+            d = max(d, _ALPHABET.index(ch) + 1)
```
refraction-ray, line 132: d should be directly given as the arg for the function instead of infer from bitsrings, to be unambiguous and consistent with other function interface with d

tensorcircuit/results/counts.py
```diff
@@ -105,44 +112,90 @@ def count2vec(count: ct, normalization: bool = True) -> Tensor:
...
+        upper = int(np.sqrt(D)) + 1
+        for d_try in range(2, max(upper, 3)):
+            try:
+                n_try = _infer_num_sites(D, d_try)
+            except ValueError:
+                continue
+            d, n = d_try, n_try
+            break
+        if d is None:
+            d, n = D, 1
```
refraction-ray, line 192: again d should be given as function input to avoid those complex infer code for d

tensorcircuit/results/counts.py
```diff
@@ -91,7 +92,13 @@ def marginal_count(count: ct, keep_list: Sequence[int]) -> ct:
 
 def count2vec(count: ct, normalization: bool = True) -> Tensor:
     """
-    Convert count dictionary to probability vector.
```
refraction-ray, line 94: add some tests functions for these qudit based count conversions

tensorcircuit/circuit.py
```diff
@@ -1,5 +1,7 @@
 """
-Quantum circuit: the state simulator
+Quantum circuit: the state simulator.
+Supports qubit (d=2) and qudit (3 <= d <= 36) systems.
+ For string-encoded samples/counts, digits use 0–9A–Z where A=10, …, Z=35.
```
refraction-ray, line 4: an extra space at the beginning

tensorcircuit/circuit.py
```diff
@@ -814,15 +818,13 @@ def measure_reference(
...
-            r = backend.real(backend.cast(r, dtypestr))
-            if r < backend.real(pu):
-                sample += "0"
-                p = p * pu
-            else:
-                sample += "1"
-                p = p * (1 - pu)
+            probs = backend.real(backend.diagonal(rho))
+            probs /= backend.sum(probs)
+            outcome = np.random.choice(self._d, p=probs)
```
refraction-ray, line 823: backend random choice?

tensorcircuit/circuit.py
```diff
@@ -916,12 +922,14 @@ def expectation(
 def expectation(
     *ops: Tuple[tn.Node, List[int]],
     ket: Tensor,
+    d: Optional[int] = None,
```
refraction-ray, line 925: better d after bra

refraction-ray: also be consistent with the var name for different function across tc, d or dim?

tensorcircuit/quantum.py
```diff
@@ -2432,14 +2469,18 @@ def partial_transpose(rho: Tensor, transposed_sites: List[int]) -> Tensor:
...
+) -> Tensor:
     """
     _summary_
 
     :param rho: _description_
     :type rho: Tensor
     :param transposed_sites: _description_
     :type transposed_sites: List[int]
+    :param d: dimension of qudit system
+    :type d: int
```
refraction-ray, line 2483: d is forgotten to be utilized in this function

tensorcircuit/quantum.py
```diff
@@ -2861,60 +2940,70 @@ def measurement_counts(
...
-        return sample2all(raw_counts, n, format=format, jittable=jittable)
+        return sample2all(raw_counts, n, format=format, jittable=jittable, d=local_d)
 
 
 measurement_results = measurement_counts
 
 
 @partial(arg_alias, alias_dict={"format": ["format_"]})
 def sample2all(
-    sample: Tensor, n: int, format: str = "count_vector", jittable: bool = False
```
refraction-ray, line 2872: add some tests for d>2 for these format conversions

tensorcircuit/quantum.py
```diff
@@ -2934,67 +3023,109 @@ def spin_by_basis(n: int, m: int, elements: Tuple[int, int] = (1, -1)) -> Tensor
...
-        r *= results[:, i]
+        col = backend.cast(results[:, i], "int32")
+        r *= backend.gather1d(evec, col)
+
     r = backend.cast(r, rdtypestr)
     return backend.mean(r)
 
 
-def correlation_from_counts(index: Sequence[int], results: Tensor) -> Tensor:
+def correlation_from_counts(
```
refraction-ray, line 3086: add tests for these correlation functions for d>2

refraction-ray (changes requested):

tensorcircuit/basecircuit.py
```diff
@@ -400,40 +429,79 @@ def measure_jit(
...
-            sign = backend.convert_to_tensor(sign)
-            sign = backend.cast(sign, dtype=rdtypestr)
-            sign_complex = backend.cast(sign, dtypestr)
-            sample.append(sign_complex)
-            p = p * (pu * (-1) ** sign + sign)
-
-        sample = backend.stack(sample)
-        sample = backend.real(sample)
+                zero_r = backend.cast(backend.convert_to_tensor(0.0), rdtypestr)
+                tiny_r = backend.cast(backend.convert_to_tensor(1e-12), rdtypestr)
```
refraction-ray, line 482: do we need a more irregular number to avoid accidental matching here?

refraction-ray:
please add some tests for these changed functions with d>2, some may require d>10, also be consistent with the variable name: d or dim

tensorcircuit/circuit.py
WeiguoMa: Given that the current measure_reference() function itself is not jittable, the implementation paths of random.choice differ in complexity across different backends, and the current abstract backend does not provide this method, can we consider directly implementing it based on simplicity and ease of use: outcome = np.random.choice(self._d, p=backend.cast(probs, dtype='float64'))? I think this is a simple and reasonable approach.

tensorcircuit/results/counts.py
WeiguoMa: Certainly

tensorcircuit/results/counts.py
WeiguoMa: Certainly

tensorcircuit/results/counts.py
WeiguoMa: great, ill just add the corresponding tests in test_counts.py.

tensorcircuit/quantum.py
WeiguoMa: This bug is fixed.

tensorcircuit/quantum.py
WeiguoMa: ok

tensorcircuit/basecircuit.py
```diff
@@ -400,40 +429,79 @@ def measure_jit(
...
+                    g2.id = id(g2)
+                    g2.is_dagger = True
+                    g2.flag = "measurement"
+                    newnodes.append(g2)
+                    g2.get_edge(0) ^ edge2[index[i]]
+                else:
+                    vec = backend.one_hot(backend.cast(sample[i], "int32"), self._d)
+                    v = backend.cast(vec, dtypestr)
+                    m = backend.tensordot(v, v, axes=0)
+                    g = Gate(m)
```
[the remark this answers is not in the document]
WeiguoMa: I'll change this to make them consistent.

tensorcircuit/circuit.py
refraction-ray: okay, cool, the users don't expect to use the reference method anyway

refraction-ray: but actually we have the random choice function `tc.backend.implicit_randc` or `stateful_randc`

tensorcircuit/abstractcircuit.py
```diff
@@ -68,6 +68,8 @@
 
 class AbstractCircuit:
     _nqubits: int
+    _d: int = 2
+    _qudit: bool = False
```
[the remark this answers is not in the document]
WeiguoMa: In QuditCircuit, we need to support the get_quoperator() and matrix() methods just like Circuit, but these two methods involve recreating Circuit instances.

Considering that we don't want users to directly use systems with dim!=2 in tc.Circuit, we need to raise an error at this time and inform the user to use tc.QuditCircuit. Considering that we are essentially wrapping tc.Circuit in tc.QuditCircuit, and also allowing users to call get_quoperator() and matrix() in QuditCircuit, it's difficult to determine whether the user is creating a new Circuit with dim!=2 or calling the two methods mentioned earlier in QuditCircuit solely through self._d. Therefore, a hidden variable self._qudit is added.

tensorcircuit/basecircuit.py
```diff
@@ -399,40 +428,74 @@ def measure_jit(
...
-            sign_complex = backend.cast(sign, dtypestr)
-            sample.append(sign_complex)
-            p = p * (pu * (-1) ** sign + sign)
-
-        sample = backend.stack(sample)
-        sample = backend.real(sample)
+                zero_r = backend.cast(backend.convert_to_tensor(0.0), rdtypestr)
+                tiny_r = backend.cast(backend.convert_to_tensor(1e-12), rdtypestr)
+                pu = backend.clip(backend.real(backend.diagonal(rho)), zero_r, one_r)
+                phi = backend.cast(
```
refraction-ray, line 476: for the implementation of cumsum, the tiny_r is unnecesary, and actually the implementation can be achieved by a simple backend.stateful_randc ?

tensorcircuit/circuit.py
```diff
@@ -42,46 +44,52 @@ class Circuit(BaseCircuit):
...
+            instead of :math:`\vert 0 \rangle^n` qubits, defaults to None.
         :type inputs: Optional[Tensor], optional
         :param mps_inputs: QuVector for a MPS like initial wavefunction.
         :type mps_inputs: Optional[QuOperator]
         :param split: dict if two qubit gate is ready for split, including parameters for at least one of
             ``max_singular_values`` and ``max_truncation_err``.
         :type split: Optional[Dict[str, Any]]
         """
+        self._d = 2 if dim is None else dim
         self.inputs = inputs
```
refraction-ray, line 70: no use of _qudit

tensorcircuit/mpscircuit.py
```diff
@@ -1015,20 +1051,26 @@ def measure(
...
-            p = p * (pu * (-1) ** sign + sign)
-            m = (1 - sign_complex) * up + sign_complex * down
+
+            cdf = backend.cumsum(ps)
+            choice = backend.sum(backend.cast(r >= cdf, "int32"))
+
+            choice_f = backend.cast(choice, dtypestr)
+            sample.append(choice_f)
+
+            m = backend.zeros((ps.shape[0],), dtype=dtypestr)
```
refraction-ray, line 1066: backend.stateful_randc

refraction-ray (changes requested):
also note how dim should always be the last args for backward compatibility

tensorcircuit/mpscircuit.py
```diff
@@ -370,42 +384,57 @@ def gate_to_MPO(
...
+        if not all(index[i] < index[i + 1] for i in range(len(index) - 1)):
+            raise AssertionError("`index` must be strictly increasing.")
+
+        index_left = int(np.min(index))
         if isinstance(gate, tn.Node):
             gate = backend.copy(gate.tensor)
-        index = np.array(index) - index_left
+
         nindex = len(index)
-        # transform gate from (in1, in2, ..., out1, out2 ...) to
```
refraction-ray, line 379: better keep this line of comment so that we know what happened in the code?

tensorcircuit/mpscircuit.py
```diff
@@ -370,42 +384,57 @@ def gate_to_MPO(
...
-            # insert identites in the middle
+        in_dims = tuple(backend.shape_tuple(gate))[:nindex]
+        dim = int(in_dims[0])
+        dim_phys_mpo = dim * dim
+
+        order = tuple(np.arange(2 * nindex).reshape(2, nindex).T.flatten().tolist())
+        gate = backend.transpose(gate, order)
+
+        index_arr = np.array(index, dtype=int) - index_left
+        pair_order = np.argsort(index_arr)
```
refraction-ray, line 405: no need to sort as you have assert the increasing order before?

WeiguoMa: Actually, the original author used argsort = np.argsort(index) here, and I'm just keeping the style. Do I need to modify it based on the actual situation?

refraction-ray: certainly, you can skip the unnecessary parts of the original implementation

tensorcircuit/results/counts.py
```diff
@@ -105,44 +116,50 @@ def count2vec(count: ct, normalization: bool = True) -> Tensor:
...
         vec = np.array(vec)
-    n = int(np.log(vec.shape[0]) / np.log(2) + 1e-9)
-    c = count_vector2dict(vec, n, key="bin")
-    if prune is True:
-        nc = c.copy()
-        for k, v in c.items():
-            if np.abs(v) < 1e-8:
-                del nc[k]
-        return nc
+    n = int(np.log(int(vec.shape[0])) / np.log(dim) + 1e-9)
```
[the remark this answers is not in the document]
WeiguoMa: Done

tensorcircuit/quantum.py
```diff
@@ -2648,117 +2704,141 @@ def count_d2s(drepr: Tensor, eps: float = 1e-7) -> Tuple[Tensor, Tensor]:
...
+        return out_int
     else:
-        dn = {}
-        for k, v in d.items():
-            kn = str(bin(k))[2:].zfill(n)
-            dn[kn] = v
-        return dn
+        out_str = {}
+        for k, v in out_int.items():
+            kn = np.base_repr(k, base=dim).zfill(n)
```
refraction-ray, line 2806: this is consitent with ABCD representation?

refraction-ray: just a check question, no potential issue

tensorcircuit/results/counts.py
```diff
@@ -105,44 +116,50 @@ def count2vec(count: ct, normalization: bool = True) -> Tensor:
...
-    nqubit = len(list(count.keys())[0])
-    probability = [0] * 2**nqubit
-    shots = sum([v for k, v in count.items()])
-    for k, v in count.items():
-        if normalization is True:
-            v /= shots  # type: ignore
-        probability[int(k, 2)] = v
-    return np.array(probability)
 
+    def parse_key(_k: str) -> List[int]:
```
refraction-ray, line 120: this is just _decode_basis_label?

tensorcircuit/basecircuit.py
```diff
@@ -399,40 +406,71 @@ def measure_jit(
...
+                        shape=1,
+                        p=backend.cast(pu, rdtypestr),
+                    )
+                else:
+                    one_r = backend.cast(backend.convert_to_tensor(1.0), rdtypestr)
+                    st = backend.cast(status[k : k + 1], rdtypestr)
+                    ind = backend.probability_sample(
+                        shots=1,
+                        p=backend.cast(pu, rdtypestr),
+                        status=one_r - st,
```
refraction-ray, line 468: why one_r-st instead of st?

tensorcircuit/circuit.py
```diff
@@ -800,10 +798,9 @@ def measure_reference(
                 if i != j:
                     e ^ edge2[i]
             for i in range(len(sample)):
-                if sample[i] == "0":
-                    m = np.array([1, 0], dtype=npdtype)
-                else:
-                    m = np.array([0, 1], dtype=npdtype)
+                m = np.array([0 for _ in range(self._d)], dtype=npdtype)
+                m[int(sample[i])] = 1
```
refraction-ray, line 802: this is also the onehot_d_tensor gate?

refraction-ray (changes requested):
In general looks nice, the pr can be merged after fixing the following minor points

tensorcircuit/results/counts.py
WeiguoMa: Yes, thank you for your patience in reviewing. Since the time of code writing varies, there may be cases where functions are not reused.

tensorcircuit/basecircuit.py
WeiguoMa: Because probability_sample internally uses 1 - status when mapping uniform random numbers to the CDF, passing one_r - st ensures the sampling behavior stays consistent with the original implicit_randc/CDF logic.

WeiguoMa:
Great, it seems that the program has passed the CI.

PR #45, opened 2025-09-03
Qudit_interface

WeiguoMa:
This PR makes following changes:
1. quditcircuit.py, this file contains tc.QuditCircuit for establishing a quadit circuit.
2. quditgates.py, this file provides basic gates in qudit systems.
3. tests/test_quditcircuit.py and tests/test_quditgates.py.
4. add interface to tc.__init__ file.

### DIFF
--- a/tensorcircuit/quditcircuit.py
+++ b/tensorcircuit/quditcircuit.py
@@ -76,7 +76,33 @@ def _cached_matrix(
     key: Optional[tuple] = (),
 ) -> Tensor:
     """
-    kind: "single" or "two"; key is a parameter tuple sorted by signature for caching
+    Build and cache a matrix using a registered builder function.
+
+    This function looks up a builder from either `SINGLE_BUILDERS` or
+    `TWO_BUILDERS` (depending on `kind`), and calls it with the given
+    arguments. Results are cached with `functools.lru_cache`, so repeated
+    calls with the same inputs return the cached tensor instead of
+    rebuilding it.
+
+    Args:
+        kind: Either `"single"` (use `SINGLE_BUILDERS`) or `"two"` (use `TWO_BUILDERS`).
+        name: The builder name to look up in the chosen dictionary.
+        d: The dimension of the matrix.
+        omega: Optional frequency or scaling parameter, passed to the builder.
+        key: Tuple of extra parameters, matched in order to the builder’s
+            expected signature.
+
+    Returns:
+        Tensor: The matrix built by the selected builder.
+
+    Notes:
+        - The cache key depends on all arguments (`kind`, `name`, `d`, `omega`, `key`).
+        - The `key` tuple must have the same order as the builder’s signature.
+        - The same inputs will always return the same cached tensor.
+
+    Raises:
+        KeyError: If the builder `name` is not found.
+        TypeError/ValueError: If `key` does not match the builder’s expected parameters.
     """
     builders = SINGLE_BUILDERS if kind == "single" else TWO_BUILDERS
     _, builder = builders[name]
@@ -146,6 +172,26 @@ class QuditCircuit:
         return self._nqubits
 
     def _apply_gate(self, *indices, name: str, **kwargs):
+        """
+        Apply a quantum gate (unitary) to one or two qudits in the circuit.
+
+        The gate matrix is looked up by name in either `SINGLE_BUILDERS` (for
+        single-qudit gates) or `TWO_BUILDERS` (for two-qudit gates). The matrix
+        is built (and cached) via `_cached_matrix`, then applied to the circuit
+        at the given indices.
+
+        Args:
+            *indices: The qudit indices the gate should act on.
+                - One index → single-qudit gate.
+                - Two indices → two-qudit gate.
+            name: The name of the gate (must exist in the chosen builder set).
+            **kwargs: Extra parameters for the gate. These are matched against
+                the gate’s signature from the builder definition.
+
+        Raises:
+            ValueError: If `name` is not found, or if the number of indices
+                does not match the gate type (single vs two).
+        """
         if len(indices) == 1 and name in SINGLE_BUILDERS:
             sig, _ = SINGLE_BUILDERS[name]
             key = tuple(kwargs.get(k) for k in sig if k != "none")
@@ -166,37 +212,102 @@ class QuditCircuit:
             raise ValueError(f"Unsupported gate/arity: {name} on {len(indices)} qudits")
 
     def any(self, *indices: int, unitary: Tensor, name: str = "any") -> None:
-        self._circ.unitary(*indices, unitary=unitary, name=name, dim=self._d)
+        self._circ.unitary(*indices, unitary=unitary, name=name, dim=self._d)  # type: ignore
 
     unitary = any
 
     def i(self, index: int) -> None:
+        """
+        Apply the identity (I) gate on the given qudit index.
+
+        Args:
+            index: Qudit index to apply the gate on.
+        """
         self._apply_gate(index, name="I")
 
     def x(self, index: int) -> None:
+        """
+        Apply the X gate on the given qudit index.
+
+        Args:
+            index: Qudit index to apply the gate on.
+        """
         self._apply_gate(index, name="X")
 
     def y(self, index: int) -> None:
+        """
+        Apply the Y gate on the given qudit index.
+
+        Args:
+            index: Qudit index to apply the gate on.
+        """
         self._apply_gate(index, name="Y")
 
     def z(self, index: int) -> None:
+        """
+        Apply the Z gate on the given qudit index.
+
+        Args:
+            index: Qudit index to apply the gate on.
+        """
         self._apply_gate(index, name="Z")
 
     def h(self, index: int) -> None:
+        """
+        Apply the Hadamard-like (H) gate on the given qudit index.
+
+        Args:
+            index: Qudit index to apply the gate on.
+        """
         self._apply_gate(index, name="H")
 
     def u8(
         self, index: int, gamma: float = 2.0, z: float = 1.0, eps: float = 0.0
     ) -> None:
+        """
+        Apply the U8 parameterized single-qudit gate.
+
+        Args:
+            index: Qudit index to apply the gate on.
+            gamma: Gate parameter gamma (default 2.0).
+            z: Gate parameter z (default 1.0).
+            eps: Gate parameter eps (default 0.0).
+        """
         self._apply_gate(index, name="U8", extra=(gamma, z, eps))
 
     def rx(self, index: int, theta: float, j: int = 0, k: int = 1):
+        """
+        Apply the single-qudit RX rotation on `index`.
+
+        Args:
+            index: Qudit index to apply the gate on.
+            theta: Rotation angle.
+            j: Source level of the rotation subspace (default 0).
+            k: Target level of the rotation subspace (default 1).
+        """
         self._apply_gate(index, name="RX", theta=theta, j=j, k=k)
 
     def ry(self, index: int, theta: float, j: int = 0, k: int = 1):
+        """
+        Apply the single-qudit RY rotation on `index`.
+
+        Args:
+            index: Qudit index to apply the gate on.
+            theta: Rotation angle.
+            j: Source level of the rotation subspace (default 0).
+            k: Target level of the rotation subspace (default 1).
+        """
         self._apply_gate(index, name="RY", theta=theta, j=j, k=k)
 
     def rz(self, index: int, theta: float, j: int = 0):
+        """
+        Apply the single-qudit RZ rotation on `index`.
+
+        Args:
+            index: Qudit index to apply the gate on.
+            theta: Rotation angle around Z.
+            j: Level where the phase rotation is applied (default 0).
+        """
         self._apply_gate(index, name="RZ", theta=theta, j=j)
 
     def rxx(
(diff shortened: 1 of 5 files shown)


tensorcircuit/quditgates.py
```diff
@@ -0,0 +1,563 @@
...
+    The Y gate represents a combination of the X and Z gates, generalizing the Pauli-Y gate
+    from qubits to higher dimensions. It is defined as
+
+    .. math::
+
+            Y = \frac{1}{i}\, Z \cdot X,
+
+    where the generalized Pauli-X and Pauli-Z gates are applied to the target qudits.
+    """
+    return np.matmul(_z_matrix_func(d, omega=omega), _x_matrix_func(d)) / 1j
```
refraction-ray, line 157: is this a well established notion? as for qudits, X^a Z^b are all similar to Y

tensorcircuit/quditgates.py
```diff
@@ -0,0 +1,563 @@
...
+def _h_matrix_func(d: int, omega: Optional[float] = None) -> Tensor:
+    r"""
+    H_d\ket{j} = \frac{1}{\sqrt{d}}\sum_{k=0}^{d-1}\omega^{jk}\ket{k}
+    """
+    omega = np.exp(2j * np.pi / d) if omega is None else omega
+    matrix = np.zeros((d, d), dtype=npdtype)
+    for j in range(d):
+        for k in range(d):
+            matrix[j, k] = omega ** (j * k) / np.sqrt(d)
+    return matrix.T
```
refraction-ray, line 169: why transpose, why not direct assign matrix[k, j]?

tensorcircuit/quditgates.py
```diff
@@ -0,0 +1,563 @@
...
+
+
+def _s_matrix_func(d: int, omega: Optional[float] = None) -> Tensor:
+    r"""
+    S_d\ket{j} = \omega^{j(j + p_d) / 2}\ket{j}
+    """
+    omega = np.exp(2j * np.pi / d) if omega is None else omega
+    _pd = 0 if d % 2 == 0 else 1
+    matrix = np.zeros((d, d), dtype=complex)
+    for j in range(d):
```
refraction-ray, line 179: np.diag

tensorcircuit/quditgates.py
```diff
@@ -0,0 +1,563 @@
...
+
+def _u8_matrix_func(
+    d: int,
+    gamma: float = 2.0,
+    z: float = 1.0,
+    eps: float = 0.0,
+    omega: Optional[float] = None,
+) -> Tensor:
+    r"""
+    U8 diagonal single-qudit gate for prime dimensions.
```
refraction-ray, line 402: add in docstring why this gate is special to be included?

tests/test_quditcircuit.py
```diff
@@ -0,0 +1,342 @@
...
+    np.testing.assert_allclose(
+        w,
+        np.array(
+            [
+                1,
+            ]
+            * 10
+        )
+        / np.sqrt(10),
+        atol=1e-4,
```
refraction-ray, line 108: at least 1e-5?

tests/test_quditgates.py
```diff
@@ -0,0 +1,295 @@
+# tests/test_quditgates.py
```
refraction-ray, line 1: delete this line

tests/test_quditgates.py
```diff
@@ -0,0 +1,295 @@
...
+    _u8_matrix_func,
+    _cphase_matrix_func,
+    _csum_matrix_func,
+    _cached_matrix,
+    SINGLE_BUILDERS,
+    TWO_BUILDERS,
+    npdtype,
+)
+
+if npdtype in (np.complex64, np.float32):
```
refraction-ray, line 27: dont use this switch in test files, use `highp` fixture instead

tests/test_quditgates.py
```diff
@@ -0,0 +1,295 @@
...
+    _swap_matrix_func,
+    _rzz_matrix_func,
+    _rxx_matrix_func,
+    _u8_matrix_func,
+    _cphase_matrix_func,
+    _csum_matrix_func,
+    _cached_matrix,
+    SINGLE_BUILDERS,
+    TWO_BUILDERS,
+    npdtype,
```
refraction-ray, line 24: no need to use the imported dtype

tensorcircuit/quditcircuit.py
```diff
@@ -0,0 +1,530 @@
...
+
+from .utils import arg_alias
+from .basecircuit import BaseCircuit
+from .circuit import Circuit
+from .quantum import QuOperator, QuVector
+from .quditgates import SINGLE_BUILDERS, TWO_BUILDERS, _cached_matrix
+
+
+Tensor = Any
+SAMPLE_FORMAT = Literal["sample_bin", "count_vector", "count_dict_bin"]
```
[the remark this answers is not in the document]
WeiguoMa: you are right, and this is an early code before Issue#41

tensorcircuit/quditgates.py
WeiguoMa: For qudits (d>2) there is no standard Pauli-Y; the generalized Pauli group is usually expressed as X^a Z^b. My current implementation defines Y \propto ZX so it reduces to the usual qubit Y. Do you prefer I keep this with a clarifying docstring, or remove Y entirely and only expose the general X^a Z^b form?

tensorcircuit/quditcircuit.py
```diff
@@ -0,0 +1,530 @@
...
+
+        Args:
+            index: Qudit index to apply the gate on.
+            theta: Rotation angle.
+            j: Source level of the rotation subspace (default 0).
+            k: Target level of the rotation subspace (default 1).
+        """
+        self._apply_gate(index, name="RY", theta=theta, j=j, k=k)
+
+    def rz(self, index: int, theta: float, j: int = 0) -> None:
```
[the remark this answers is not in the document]
WeiguoMa: In the qudit literature both appear: subspace rotations (our current RX/RY/RZ on levels j,k) are natural in experimental settings where one drives a two-level transition, while global rotations defined from exp(i\theta(X+X^\dagger)/2) or exp(i\theta Z) are more common in algebraic/Clifford–Weyl formulations. We can keep the subspace version for physical relevance, but can also add the global form if you prefer.

tensorcircuit/quditgates.py
refraction-ray: it depends on whether Y gate in qudit case is a well established notion, if the common notion is that there is no exact Y gate counterpart for qudits, then we can skip the Y gate

tests/test_quditgates.py
```diff
@@ -0,0 +1,290 @@
...
+    A2 = _cached_matrix("single", "I", d=3, omega=None, key=())
+    assert A1 is A2
+
+    X1 = _cached_matrix("single", "RX", d=3, omega=None, key=(0.1, 0, 1))
+    X2 = _cached_matrix("single", "RX", d=3, omega=None, key=(0.2, 0, 1))
+    assert X1 is not X2
+    assert X1.shape == (3, 3) and X2.shape == (3, 3)
+
+
+def test_builders_smoke(highp):
```
refraction-ray, line 271: what does smoke mean?

tensorcircuit/quditgates.py
```diff
@@ -0,0 +1,554 @@
...
+    :rtype: Tensor
+    """
+    D = d * d
+    M = np.eye(D, dtype=npdtype)
+
+    # flatten basis index: |a,b> ↦ a*d + b
+    idx_a = j1 * d + j2
+    idx_b = k1 * d + k2
+
+    c = np.cos(theta / 2.0)
```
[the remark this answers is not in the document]
WeiguoMa: i'll change all the formation of variable gates in next push

refraction-ray (changes requested):

examples/vqe_qudit_example.py
```diff
@@ -0,0 +1,271 @@
...
+  - Minimizes the expectation of a simple 2-site Hermitian Hamiltonian:
+        H = N(0) + N(1) + J * [ X_sym(0)\otimes X_sym(1) + Z_sym(0)\otimes Z_sym(1) ]
+    where N = diag(0,1,...,d-1), X_sym = (X + X^\dagger)/2, Z_sym = (Z + Z^\dagger)/2.
+"""
+
+import argparse
+import math
+import sys
+from dataclasses import dataclass
+from typing import List, Sequence, Tuple
```
refraction-ray, line 23: no need to add type annotation for examples, though it is okay to keep

examples/vqe_qudit_example.py
```diff
@@ -0,0 +1,271 @@
...
+import numpy as np
+import tensorcircuit as tc
+from tensorcircuit.quditcircuit import QuditCircuit
+
+
+# ---------- Hamiltonian helpers ----------
+
+
+def number_op(d: int) -> np.ndarray:
+    return np.diag(np.arange(d, dtype=np.float32)).astype(np.complex64)
```
refraction-ray, line 34: what about use set_dtype, and all the dtype used follow the convention of tc.set_dtype?

examples/vqe_qudit_example.py
```diff
@@ -0,0 +1,271 @@
...
+
+
+def x_unitary(d: int) -> np.ndarray:
+    X = np.zeros((d, d), dtype=np.complex64)
+    for j in range(d):
+        X[(j + 1) % d, j] = 1.0
+    return X
```
refraction-ray, line 41: for x and z gate, what about directly call gate matrix from quditgates.py, and for N gate it is just a shift of Z gate?

examples/vqe_qudit_example.py
```diff
@@ -0,0 +1,271 @@
...
+def symmetrize_hermitian(U: np.ndarray) -> np.ndarray:
+    return 0.5 * (U + U.conj().T)
+
+
+def kron2(a: np.ndarray, b: np.ndarray) -> np.ndarray:
+    return np.kron(a, b).astype(np.complex64)
+
+
+@dataclass
+class Hamiltonian2Qudit:
```
refraction-ray, line 59: no need to define this extra layer of abstract in the example?

examples/vqe_qudit_example.py
```diff
@@ -0,0 +1,271 @@
...
+    return best_x, float(best_y)
+
+
+def gradient_descent_ad(energy_bk, x0_np: np.ndarray, steps=200, lr=0.1, jit=False):
+    """
+    energy_bk: (backend_tensor[nparams]) -> backend_scalar
+    Simple gradient descent in numpy space with backend-gradients.
+    """
+    bk = tc.backend
+    if jit and hasattr(bk, "jit"):
```
refraction-ray, line 175: all backend has jit, even for numpy, though it fallbacks to do nothing

tests/test_quditcircuit.py
```diff
@@ -0,0 +1,446 @@
...
+    thetas = tc.array_to_tensor(np.linspace(-1.0, 1.0, 7))
+    vvals = venergy(thetas)
+    eager_vals = np.array([energy(t) for t in thetas])
+
+    np.testing.assert_allclose(
+        tc.backend.numpy(vvals), eager_vals, rtol=1e-6, atol=1e-6
+    )
+
+
+def test_quditcircuit_single_and_two_qudit_paths_and_wrappers():
```
refraction-ray, line 429: is this function name proper for what the test does?

tests/test_quditcircuit.py
```diff
@@ -0,0 +1,446 @@
...
+    for bad in ["sample_int", "count_tuple", "count_dict_int", "count_vector"]:
+        with pytest.raises(NotImplementedError):
+            c.sample(allow_state=False, batch=1, format=bad)
+
+
+def test_quditcircuit_amplitude_before_wrapper():
+    c = tc.QuditCircuit(2, 3)
+    c.x(0)
+    nodes = c.amplitude_before("00")
+    assert isinstance(nodes, list)
```
refraction-ray, line 446: also check the length of the list (number of nodes?)

tests/test_quditvqe.py
```diff
@@ -0,0 +1,72 @@
+# tests/test_quditvqe.py
```
refraction-ray, line 1: please remove this test file

refraction-ray (changes requested):
The general implementation now looks good to me, some minor changes required to minimize and clean up the qudit vqe example

examples/vqe_qudit_example.py
WeiguoMa: This code is generally checked and fixed by GPT, it is fine to keep it.

examples/vqe_qudit_example.py
WeiguoMa: fine for this

examples/vqe_qudit_example.py
WeiguoMa: I found that there is no need to specify any dtype, it is fine to keep it as default. So I removed everything that is relevant to this.

examples/vqe_qudit_example.py
WeiguoMa: Using X and Z directly from quditgates is fine; we're just building a NumPy matrix in this process. What is the N gate? N = np.diag(np.arange(d))? I think it's fine to use it this way, and I've already optimized the code to make it more concise, so I don't think it's necessary to change this part anymore.

refraction-ray (changes requested):
Some more fixes required to make the example script more clean, imagine using jax backend on GPU, you will not like the conversion to numpy happen during the computation as it requires CPU-GPU data transfer which is slow, so try to avoid such conversion unless it only happens at the very beginning stage

WeiguoMa:
Hi, Shixin. I tried to make a ``new'' file here. This should be totally different and cleaner without any complicated conversion.

refraction-ray:
nice work, LGTM!

PR #48, opened 2025-09-09
Bug fix in html docsing

WeiguoMa:
making quditgate functions into pulic (by removing '_'), this fix the bug in docsing.


WeiguoMa:
please wait to merge this PR, I will check the docsing in quditcircuit.py (which should be done in next PR) tommorow, and add a commit here.

PR #57, opened 2026-01-29
Add example: simulation for measurement-only circuits 

iop-hui252:
This script simulates a quantum circuit under monitoring (weak measurements). It calculates the evolution of a quantum state under weak X, ZZ, and ZXZ measurements. It also computes topological entanglement entropy and mutual Information.


refraction-ray (approved):
LGTM

PR #82, opened 2026-02-26
Support hyperedges in TensorCircuit with cotengra

refraction-ray:
Implemented hyperedge support for cotengra contractor by handling CopyNodes as hyperedges in path finding and contraction execution.

---
*PR created automatically by Jules for task [16871959254681859317](https://jules.google.com/task/16871959254681859317) started by @refraction-ray*

### DIFF
--- a/tensorcircuit/cons.py
+++ b/tensorcircuit/cons.py
@@ -517,7 +517,7 @@ def _identity(*args: Any, **kws: Any) -> Any:
     return args
 
 
-def _get_path_cache_friendly(
+def _get_path_info(
     nodes: List[tn.Node], algorithm: Any
 ) -> Tuple[List[Tuple[int, int]], List[tn.Node], Dict[Any, str]]:
     # Refactored to return output_symbols as well
@@ -622,20 +622,12 @@ def _get_path_cache_friendly(
 
     size_dict = {}
     for root, symbol in mapping_dict.items():
-        size_dict[symbol] = root.dimension  # type: ignore  # type: ignore
+        size_dict[symbol] = root.dimension  # type: ignore
 
     logger.debug("input_sets: %s" % input_sets)
     logger.debug("output_set: %s" % output_set)
     logger.debug("size_dict: %s" % size_dict)
     logger.debug("path finder algorithm: %s" % algorithm)
-
-    # We need a way to map original edges to symbols for rebind_dangling_edges.
-    # uf[edge] -> root -> symbol (in mapping_dict)
-    # So we return uf and mapping_dict? Or just a helper dict.
-    # Let's return a dict {edge: symbol} for all edges?
-    # Or better, just return the `uf` and `mapping_dict`.
-    # But `uf` is not picklable or standard?
-    # Let's construct a simple dict: {id(edge): symbol}
 
     edge_to_symbol = {}
     for edge in all_edges:
@@ -647,6 +639,14 @@ def _get_path_cache_friendly(
         regular_nodes,
         edge_to_symbol,  # Use this instead of output_set/mapping_dict for output reordering
     )
+
+
+def _get_path_cache_friendly(
+    nodes: List[tn.Node], algorithm: Any
+) -> Tuple[List[Tuple[int, int]], List[tn.Node]]:
+    # Legacy wrapper for backward compatibility of get_tn_info
+    path, regular_nodes, _ = _get_path_info(nodes, algorithm)
+    return path, regular_nodes
 
 
 get_tn_info = partial(_get_path_cache_friendly, algorithm=_identity)
@@ -760,7 +760,7 @@ def _base(
     # nodes = list(nodes_set)
 
     # 1. FRONTEND: Resolve topology
-    path, regular_nodes, edge_to_symbol = _get_path_cache_friendly(nodes, algorithm)
+    path, regular_nodes, edge_to_symbol = _get_path_info(nodes, algorithm)
 
     # Detect if we should use the new primitive-based engine
     # If the number of regular nodes returned differs from input nodes (meaning CopyNodes were filtered out),
@@ -826,15 +826,11 @@ def _base(
     # ==========================================
     # be = regular_nodes[0].backend
     be = backend
-
     # Determine output symbols
     output_symbols = set()
-    if output_edge_order:
-        for edge in output_edge_order:
-            # We must use edge_to_symbol
-            # If edge was not in edge_to_symbol (e.g. not connected?), this would key error.
-            # But edge_to_symbol covers all edges in nodes_new.
-            # Since output_edge_order edges are dangling edges of the graph, they are in all_edges.
+    dangling_edges = tn.get_subgraph_dangling(nodes)
+    for edge in dangling_edges:
+        if id(edge) in edge_to_symbol:
             output_symbols.add(edge_to_symbol[id(edge)])
 
     # Extract bare tensors and their initial symbols into a working pool


tests/test_hyperedge.py
```diff
@@ -0,0 +1,93 @@
...
+def test_tensorcircuit_circuit_hyperedge_support():
+    # While TC circuit doesn't typically create CopyNodes directly in gates,
+    # ensuring the contractor works with general graphs is key.
+    # This test just ensures normal circuit simulation still works with cotengra
+    # (which implies the new logic handles regular nodes correctly too).
+    c = tc.Circuit(2)
+    c.H(0)
+    c.CNOT(0, 1)
+
+    tc.set_contractor("cotengra")
```
refraction-ray: dont set contengra like this, what is the test fails, the set cannot be cleared, should define and use as a fixture

refraction-ray:
black! all the file should be blacked

refraction-ray:
Here is the complete, self-contained Code Review report. You can copy and paste this directly into the PR comments. It covers the praise for the pathfinding, the explanation of the OOM issue, the architectural shift to bare tensors with explicit primitives, and the request for a fallback toggle.📝 Code Review: Native Hyperedge (CopyNode) SupportFirst off, thank you for this PR! Tackling native hyperedge support is one of the most challenging architectural problems in tensor network frameworks.Overall Assessment: The pathfinding phase (using UnionFind) is brilliant and conceptually flawless. However, the execution phase (_base) still falls back to contracting CopyNode objects physically, which leaves a severe Out-Of-Memory (OOM) timebomb in the code. Furthermore, for the execution phase, we should shift away from physical graph objects entirely and use bare tensors with explicit backend primitives. Finally, we need to keep the legacy execution path as a safe fallback.Here is a detailed breakdown of the logic and actionable suggestions for the next iteration.✅ The Good: Flawless Pathfinding via UnionFindYour logic in _get_path_cache_friendly using UnionFind is excellent. By treating CopyNode as a virtual router rather than a physical computing node, you effectively group all connected edges into a single connected component (root). Assigning a unified symbol to these components maps the graph topology perfectly into the algebraic input_sets expected by cotengra / opt_einsum. This completely solves the hyperedge pathfinding problem!💣 The Issue: The CopyNode OOM Timebomb in ExecutionIn the _base execution loop, the code attempts to perform the contraction physically on the graph using this logic:Python        shared_cns = set()
        # ... finding shared_cns ...
        curr_node_a = node_a
        for cn in shared_cns:
            curr_node_a = tn.contract_between(curr_node_a, cn) # 🚨 DANGER
Massive Memory Allocation: A CopyNode in tensornetwork is not magical. When you force a tn.contract_between(node, cn) step by step, the underlying backend usually instantiates the CopyNode.tensor. If a CopyNode connects 20 edges (dimension $d$), it will attempt to allocate a dense $O(d^{20})$ all-zero tensor with a diagonal of ones. This will instantly trigger an OOM error.Topology Corruption: By eagerly "eating" the shared CopyNode, the uncontracted dangling edges of that CopyNode are prematurely grafted onto curr_node_a. This disrupts the global contraction path meticulously planned by the optimizer, often leading to massive unintentional outer products later in the loop.💡 Architectural Shift: Bare Tensors & Explicit PrimitivesDuring the actual contraction loop, we don't need tn.Node or tn.Edge objects at all. Continuing to wrap intermediate results in tn.Node introduces unnecessary Python object overhead and risks accidental graph mutations.Instead, we should maintain a pool of bare tensors and their corresponding string symbols. Furthermore, to retain maximum control over the computation graph (for JIT, AutoDiff, and potential SVD truncations), we should avoid using a black-box einsum and instead manually dispatch the contractions to explicit backend primitives (matmul/tensordot for standard edges, and multiply/batched operations for hyperedges).🛡️ The Fallback: Keep the Legacy ImplementationWhile this new primitive-based execution is the future, the original _base implementation (using tn.contract_between) is extensively verified and battle-tested. We must keep the original logic intact behind a switch/toggle. This allows users to safely fall back to the old behavior (which works perfectly for standard, non-hyperedge networks) and allows us to easily benchmark the performance differences between the two execution engines.🛠️ Proposed Code StructureHere is how we can implement the toggle and the bare-tensor pool approach in _base:Pythondef _base(
    nodes: List[tn.Node],
    algorithm: Any,
    output_edge_order: Optional[Sequence[tn.Edge]] = None,
    ignore_edge_order: bool = False,
    total_size: Optional[int] = None,
    debug_level: int = 0,
    use_primitives: bool = False, # 👈 NEW: Toggle for the execution engine
) -> tn.Node:
    
    # 1. FRONTEND: Resolve topology
    # (Your UnionFind logic goes here. If use_primitives is False, it can fallback to original behavior)
    path, regular_nodes, output_symbols = _get_path_cache_friendly(nodes, algorithm)
    
    if not use_primitives:
        # ==========================================
        # LEGACY EXECUTION PATH (Safe Fallback)
        # ==========================================
        # Put the ORIGINAL `for ab in path: ... tn.contract_between ...` loop here.
        # This guarantees zero breaking changes for existing code.
        ...
        return final_node

    # ==========================================
    # NEW EXECUTION PATH (Hyperedge & JIT friendly)
    # ==========================================
    be = regular_nodes[0].backend 
    
    # Extract bare tensors and their initial symbols into a working pool
    # Pool elements are just tuples: (raw_tensor, "abc")
    tensor_pool = [(node.tensor, getattr(node, "_symbols")) for node in regular_nodes]

    for ab in path:
        if len(ab) < 2:
            continue
        a, b = ab
        tensor_a, sym_a = tensor_pool[a]
        tensor_b, sym_b = tensor_pool[b]

        # Calculate remaining symbols needed by the rest of the pool
        remaining_pool = [t for i, t in enumerate(tensor_pool) if i not in (a, b)]
        symbols_left = set("".join([sym for _, sym in remaining_pool]))
        
        # Categorize shared axes
        shared_syms = set(sym_a).intersection(set(sym_b))
        trace_syms = [s for s in shared_syms if s not in symbols_left and s not in output_symbols]
        hyper_syms = [s for s in shared_syms if s in symbols_left or s in output_symbols]

        # Dispatch to BARE primitives 
        if not hyper_syms:
            # Standard tensor contraction
            new_tensor = explicit_tensordot(be, tensor_a, tensor_b, trace_syms)
        else:
            # Hyperedge involved (requires alignment, broadcast, multiply, then trace)
            new_tensor = explicit_batched_multiply(be, tensor_a, tensor_b, trace_syms, hyper_syms)

        sym_out = compute_new_symbols(sym_a, sym_b, trace_syms)
        
        # Add the BARE result back to the pool
        tensor_pool.append((new_tensor, sym_out))
        tensor_pool = _multi_remove(tensor_pool, [a, b])

    # RE-ENTRY: Wrap the final bare tensor back into the Graph world
    final_raw_tensor, final_symbols = tensor_pool[0]
    final_node = tn.Node(final_raw_tensor, backend=be)
    
    # Rebind dangling edges to final_node based on final_symbols and mapping_dict
    if not ignore_edge_order and output_edge_order is not None:
        rebind_dangling_edges(final_node, output_edge_order, final_symbols, mapping_dict)

    return final_node
Next StepsCould you update the PR to encapsulate the legacy _base loop within an if not use_primitives: block, and adopt the bare-tensor pool for the hyperedge logic?

pylint should also apply to tests/

tensorcircuit/cons.py
```diff
@@ -656,46 +728,218 @@ def _base(
...
+    # If we re-run that logic or pass the map, we can know which symbol corresponds to which original edge.
+
+    # However, `output_symbols` returned from `_get_path_cache_friendly` is a list of symbols
+    # corresponding to `sorted_edges(tn.get_subgraph_dangling(nodes))` of the ORIGINAL graph.
+
+    # So `output_symbols`[i] corresponds to the i-th edge in `sorted_edges(...)`.
+
+    # `final_symbols` is the actual axis order of `final_raw_tensor`.
+
+    # We need to permute `final_raw_tensor` so that its axes match `output_edge_order`.
```
refraction-ray: dont comment out loudly unless necessary

tensorcircuit/cons.py
```diff
@@ -656,46 +759,237 @@ def _base(
...
+
+    # Implementation strategy:
+    # 1. We need the map {original_edge: symbol}.
+    # 2. `output_edge_order` is a list of `original_edge`.
+    # 3. We find the symbol for each edge in `output_edge_order`.
+    # 4. We find the current index of that symbol in `final_symbols`.
+    # 5. We construct the permutation.
+
+    # To do this, we need `uf` and `mapping_dict` access.
+    # I will modify `_get_path_cache_friendly` to return a lookup function/dict.
```
refraction-ray: delete these comments

PR #91, opened 2026-03-08
feat: reproduce 1810.03787 fig2c using arxiv-reproduce skill

refraction-ray:
This PR implements the reproduction of Figure 2(c) from the Quantum Convolutional Neural Networks paper (arXiv:1810.03787). It follows the `.agents/skills/arxiv-reproduce` workflow. 

It generates the expected plot showing a phase transition at roughly `h2/J = 0.423`. It includes an exact calculation implementation for a smaller scale `N=11` to cross check with a large scale `N=45` implementation using DMRG from `quimb` passed into TensorCircuit.

---
*PR created automatically by Jules for task [6076547423728011896](https://jules.google.com/task/6076547423728011896) started by @refraction-ray*

### DIFF
--- a/examples/reproduce_papers/2018_qcnn/main.py
+++ b/examples/reproduce_papers/2018_qcnn/main.py
@@ -206,8 +206,8 @@ if __name__ == "__main__":
         methods=["greedy"],
         parallel=False,
         minimize="combo",
-        max_time=0.5,
-        max_repeats=8,
+        max_time=1.0,
+        max_repeats=16,
         progbar=False,
     )
     tc.set_contractor("custom", optimizer=opt, preprocessing=True)
--- a/run_format.sh
+++ b/run_format.sh
@@ -1,2 +1,0 @@
-black examples/reproduce_papers/2018_qcnn/main.py
-pylint examples/reproduce_papers/2018_qcnn/main.py


refraction-ray:
1. The Convolution Layer Missing Cross-Block CZ GatesThe Issue: In exact_qcnn_circuit, you apply the convolution layer's CZ gates only within the 3-qubit pooling blocks (q1-q2 and q2-q3). However, the paper explicitly states that the convolution layers are "translationally-invariant". In Figure 2(b), the blue CZ lines form a continuous chain connecting all adjacent qubits across the entire system. By restricting the CZ gates to isolated 3-qubit blocks, you break the MERA inversion topology and lose the entanglement between the blocks.The Fix: You need to apply the CZ gates across all adjacent current_qubits first, and then apply the 3-qubit pooling operations.Python# 1. Translationally invariant convolution layer
for i in range(len(current_qubits) - 1):
    c.cz(current_qubits[i], current_qubits[i + 1])

# 2. Pooling layer (Toffoli and conditional phase flips)
next_qubits = []
for i in range(0, len(current_qubits) - 2, 3):
    q1 = current_qubits[i]
    q2 = current_qubits[i + 1]
    q3 = current_qubits[i + 2]

    # CZs are already applied above, so we skip them here
    
    # Toffoli with controls in X basis
    c.H(q1)
    c.H(q3)
    c.toffoli(q1, q3, q2)
    c.H(q1)
    c.H(q3)
    
    # Phase flips when adjacent measurement is X=-1
    c.H(q1)
    c.cz(q1, q2)
    c.H(q1)

    c.H(q3)
    c.cz(q3, q2)
    c.H(q3)

    next_qubits.append(q2)
2. Double-Negative in the Fully Connected (FC) Layer MeasurementThe Issue: In calculate_qcnn, you apply a final sequence of CZ gates and then measure the expectation value of $Z \otimes X \otimes Z$.According to the paper, the fully connected layer "applies controlled-phase gates followed by an X projection, effectively measuring $Z_{i-1} X_i Z_{i+1}$".Mathematically, if you conjugate an X operator by CZ gates, you get $ZXZ$. Because you are applying the CZ gates and then measuring $ZXZ$, you are effectively measuring $(CZ) ZXZ (CZ) = X$ in the basis prior to the FC layer! You are accidentally undoing the FC layer's logic.The Fix: You should either apply the CZ gates and measure just the middle $X$, OR skip the CZ gates and measure $ZXZ$. To match the literal operations in Figure 2(b), applying the CZs and measuring $X$ is the most physically accurate approach:Pythondef calculate_qcnn(circ, rem_qubits):
    # Fully connected layer applies CZ gates
    for i in range(len(rem_qubits) - 1):
        circ.cz(rem_qubits[i], rem_qubits[i + 1])

    if len(rem_qubits) >= 3:
        mid_idx = len(rem_qubits) // 2
        q2 = rem_qubits[mid_idx]
        
        # Followed by an X projection
        val = circ.expectation(
            [tc.gates.x(), [q2]],
            reuse=False,
        )
        return abs(float(val.real))
        
    if len(rem_qubits) >= 1:
        mid_idx = len(rem_qubits) // 2
        val = circ.expectation_ps(x=[rem_qubits[mid_idx]], reuse=False)
        return abs(float(val.real))
    return 0.0

PR #110, opened 2026-04-16
Add QSP and QSVT modules with tests

DeanTMaxim:


refraction-ray:
please first fix these issues reported in the automated replies and the github workflow

refraction-ray:
for most tc examples, one should use jax backend instead of torch

DeanTMaxim:
Thanks for your valuable feedback! I have updated the code to address your concerns.

> please first fix these issues reported in the automated replies and the github workflow

check_all.sh
```diff
@@ -1,7 +1,7 @@
 #! /bin/sh
```
refraction-ray, line 1: no need to change this file, why this change?

examples/_optimize_helpers.py
```diff
@@ -0,0 +1,104 @@
...
+    initial_parameters: np.ndarray,
+    maxiter: int,
+) -> object:
+    if importlib.util.find_spec("jaxopt") is None:
+        raise ImportError(
+            "jax backend optimization for this example requires jaxopt. "
+            "Install jaxopt in the active environment or switch to a non-jax backend."
+        )
+
+    import jaxopt
```
refraction-ray: use optax instead jaxopt, no defensive programming from importing a package in example scripts, just import at the top of file

examples/_optimize_helpers.py
```diff
@@ -0,0 +1,104 @@
...
+import numpy as np
+from scipy.optimize import minimize
+
+import tensorcircuit as tc
+
+
+def _jit_loss(loss: Callable[[np.ndarray], object]) -> Callable[[np.ndarray], object]:
+    if tc.backend.name == "numpy":
+        return loss
+    return tc.backend.jit(loss)
```
refraction-ray: no need to branch different backend, numpy also has jit (identity)

examples/qsp.py
```diff
@@ -2,48 +2,136 @@
...
 """
 
 from typing import Callable, Optional, Union
 
 import numpy as np
-from scipy.optimize import minimize
 
 import tensorcircuit as tc
 
+if __package__:
```
refraction-ray: too defensive, the example script is never part a package, why bother?

examples/qsp.py
```diff
@@ -2,48 +2,136 @@
...
+        ],
+        axis=0,
+    )
+
+
+def _signal_rotation(x: object) -> object:
+    x_tensor = _clip_signal(_real_tensor(x))
+    x_complex = _complex_tensor(x_tensor)
+    y_tensor = tc.backend.sqrt(tc.backend.relu(1.0 - x_tensor * x_tensor))
+    off_diagonal = -_complex_tensor(1.0j) * _complex_tensor(y_tensor)
```
refraction-ray: no need to wrap 1.j

examples/qsp.py
```diff
@@ -2,48 +2,136 @@
...
+        _real_tensor(target_values),
+    )
+
+
+def _evaluate_qsp_polynomial(phases: np.ndarray, x_samples: np.ndarray) -> np.ndarray:
+    x_tensor = _real_tensor(np.asarray(x_samples, dtype=np.float64))
+    values = tc.backend.vmap(
+        lambda sample: qsp_polynomial(phases, sample), vectorized_argnums=0
+    )(x_tensor)
+    return np.asarray(tc.backend.numpy(values), dtype=np.float64)
```
refraction-ray: why return numpy

examples/qsp.py
```diff
@@ -0,0 +1,285 @@
...
+    result = minimize_backend_loss(
+        loss,
+        initial_phases,
+        maxiter=1000,
+        use_gradient=use_gradient,
+    )
+    return result
+
+
+def fit_qsp_phases_variational(
```
refraction-ray: what is the diff between the two functions?

examples/qsvt.py
```diff
@@ -0,0 +1,256 @@
+"""
+Quantum Singular Value Transformation (QSVT) example.
+
+This script contains a small QSVT solver and runs a basic identity-target
+demo when executed directly.
+"""
+
```
refraction-ray, line 7: qsvt should directly use the fitted angle from single qubit qsp instead of fitting on the multi qubit directly

examples/qsvt.py
```diff
@@ -0,0 +1,256 @@
...
+        ],
+        axis=0,
+    )
+    return tc.backend.kron(
+        ancilla_rotation,
+        tc.backend.eye(system_dimension, dtype=tc.dtypestr),
+    )
+
+
+def hermitian_sqrt(matrix: object) -> object:
```
refraction-ray: we have backend.sqrth already?

refraction-ray:
Two comments are NOT addressed:

  1. check_all.sh — change still present

  The reviewer said "no need to change this file, why this change?" but the diff still swaps black . --check → black --check .. This file should be reverted.

  2. _optimize_helpers.py — numpy backend branching remains

  The reviewer said "no need to branch different backend, numpy also has jit (identity)" but the code still has:
  if tc.backend.name == "numpy":
      # special numpy path
  else:
      loss_interface = tc.interfaces.scipy_interface(...)
  and also if tc.backend.name == "jax" in minimize_backend_loss. The reviewer's point is that tc.backend.jit on numpy is already a no-op identity, so the numpy
  special case is unnecessary.

since _optimize_helpers.py is only imported by qsp.py, and qsvt.py only needs the QSP-level fitting functions, the correct structure should be:

  qsp.py  ←  qsvt.py

  The _minimize_with_scipy, _minimize_with_optax, and minimize_backend_loss functions from _optimize_helpers.py should be moved directly into qsp.py as private
  helpers. The _optimize_helpers.py file should be deleted.

DeanTMaxim:
Thanks for the feedback. The numpy branch is kept because scipy_interface currently relies on tensorflow.nest for pytree operations (via numpy_args_to_backend), which causes failures in environments without TensorFlow (like my standard Windows setup).

examples/qsp.py
```diff
@@ -0,0 +1,379 @@
...
+    return fit_qsp_phases_variational(
+        target=lambda x: x,
+        degree=degree,
+        x_samples=x_samples,
+    )
+
+
+def main() -> None:
+    """Run the QSP classical and circuit-based demonstrations."""
+    target = lambda x: x
```
refraction-ray, line 188: better set jax backend here?

examples/qsvt.py
```diff
@@ -0,0 +1,248 @@
...
+"""
+
+from typing import Callable, Optional, Tuple
+
+import numpy as np
+
+from qsp import fit_qsp_phases, fit_qsp_phases_variational
+
+import tensorcircuit as tc
+
```
refraction-ray, line 16: set_backend("jax") by default

examples/qsvt.py
```diff
@@ -0,0 +1,248 @@
...
+    matrix_dag = tc.backend.adjoint(matrix_tensor)
+    identity = tc.backend.eye(matrix_tensor.shape[0], dtype=tc.dtypestr)
+    sqrt_right = tc.backend.sqrtmh(identity - matrix_tensor @ matrix_dag, psd=True)
+    sqrt_left = tc.backend.sqrtmh(identity - matrix_dag @ matrix_tensor, psd=True)
+    top = tc.backend.concat([matrix_tensor, sqrt_right], axis=1)
+    bottom = tc.backend.concat([sqrt_left, -matrix_dag], axis=1)
+    return tc.backend.concat([top, bottom], axis=0)
+
+
+class QSVTSolver:
```
refraction-ray: no need for a class here, pure functional is better

examples/qsvt.py
```diff
@@ -0,0 +1,248 @@
...
+            result = fit_qsp_phases(
+                self.target_function,
+                degree=self.degree,
+                x_samples=x_samples,
+                initial_phases=initial_phases,
+                maxiter=max_steps,
+            )
+            return result.x
+        if self.solver_type == "variational":
+            return fit_qsp_phases_variational(
```
refraction-ray: why bother have these two functions if they are essentially the same thing

examples/qsp.py
```diff
@@ -0,0 +1,379 @@
...
+        )
+        raw_parameters = optax.apply_updates(raw_parameters, updates)
+
+    if loss_value is None or gradients is None:
+        loss_value, gradients = value_and_grad(raw_parameters)
+
+    bounded_params = np.pi * tc.backend.tanh(raw_parameters)
+    gradient_norm = float(tc.backend.numpy(tc.backend.norm(gradients)))
+
+    class OptaxResult:
```
refraction-ray: why bother a class, just return a tuple or namedtuple?

examples/qsp.py
```diff
@@ -0,0 +1,379 @@
...
+
+
+def _prepare_qsp_fit_inputs(
+    target: Callable[[Union[float, np.ndarray]], Union[float, np.ndarray]],
+    degree: int,
+    x_samples: Optional[np.ndarray],
+    initial_phases: Optional[np.ndarray],
+) -> tuple[np.ndarray, np.ndarray, object, object]:
+    if x_samples is None:
+        x_samples = np.linspace(-1.0, 1.0, 81)
```
refraction-ray: why hard coded 81?

examples/qsp.py
```diff
@@ -0,0 +1,379 @@
...
+
+def _fit_qsp_phases(
+    target: Callable[[Union[float, np.ndarray]], Union[float, np.ndarray]],
+    degree: int,
+    evaluator: Callable[[np.ndarray, object], object],
+    x_samples: Optional[np.ndarray],
+    initial_phases: Optional[np.ndarray],
+    maxiter: int,
+) -> object:
+    _, initial_phases, x_tensor, target_tensor = _prepare_qsp_fit_inputs(
```
refraction-ray: _prepare_qsp_fit_inputs returns a value it immediately discards
  The caller does _, initial_phases, x_tensor, target_tensor = _prepare_qsp_fit_inputs(...) — the first return value x_samples is thrown away. A function that
  returns something only to be discarded has the wrong boundary. It is also only called once, so it should just be inlined.

examples/qsp.py
```diff
@@ -0,0 +1,379 @@
...
+        degree,
+        _qsp_circuit_value,
+        x_samples,
+        initial_phases,
+        maxiter,
+    )
+    return result.x
+
+
+def run_quantum_solver_example(
```
refraction-ray: Only called from main(), hardcodes target=lambda x: x, and wraps a single function call. This should be inlined in main().

examples/qsvt.py
```diff
@@ -0,0 +1,248 @@
...
+            circuit.rz(0, theta=-2.0 * phases[k])
+            if k > 0:
+                gate = self._transition_unitary(degree, k, unitary_tensor, unitary_dag)
+                if (degree + k) % 2 == 1:
+                    circuit.any(*unitary_qubits, unitary=gate, name="Udag")
+                else:
+                    circuit.any(*unitary_qubits, unitary=gate, name="U")
+        return circuit
+
+    def validate_phases(
```
refraction-ray: Removing build_qsvt_unitary / _phase_rotation_for_system / _matrix_block leaves _circuit_block intact, and the circuit error alone check is sufficient, and this script would be much consice and clear

refraction-ray (commented):
please reduce the over deisgn and over defensive pattern in the code as commented attached

examples/qsp.py
```diff
@@ -0,0 +1,295 @@
...
+    use_gradient: bool,
+) -> object:
+    optimized_loss = tc.backend.jit(loss)
+    if tc.backend.name == "numpy":
+
+        def numpy_loss(phases: np.ndarray) -> float:
+            return float(tc.backend.numpy(optimized_loss(phases)))
+
+        loss_interface = numpy_loss
+        jac = False
```
refraction-ray, line 37: if jac is False, how does LBFGS work?

examples/qsp.py
```diff
@@ -0,0 +1,295 @@
...
+        return parameters, state, loss_value, gradients
+
+    loss_value = None
+    gradients = None
+    for _ in range(maxiter):
+        raw_parameters, optimizer_state, loss_value, gradients = update_step(
+            raw_parameters, optimizer_state
+        )
+
+    if loss_value is None or gradients is None:
```
refraction-ray: what does this over defensive even do?

examples/qsp.py
```diff
@@ -0,0 +1,295 @@
...
+    return tc.backend.stack(
+        [
+            tc.backend.stack([x_complex, off_diagonal], axis=0),
+            tc.backend.stack([off_diagonal, x_complex], axis=0),
+        ],
+        axis=0,
+    )
+
+
+def _qsp_circuit_value(phases: np.ndarray, x: object) -> object:
```
refraction-ray: you are still using both unitary and circuit. no need, please follow my comment...
just remove all circuit part, for the evaluation and fit, use unitary part

examples/qsvt.py
refraction-ray: ?

examples/qsvt.py
```diff
@@ -0,0 +1,253 @@
...
+    index: int,
+    unitary: object,
+    unitary_dag: object,
+) -> object:
+    if (degree + index) % 2 == 1:
+        return unitary_dag
+    return unitary
+
+
+def build_qsvt_unitary(
```
refraction-ray: just delete all unitary part, please follow my comment...

refraction-ray (commented):
many of my comment are not followed, please read them carefully....

examples/qsp.py
```diff
@@ -0,0 +1,208 @@
...
+) -> MinimizeResult:
+    if maxiter is None:
+        maxiter = 500
+    if x_samples is None:
+        n_samples = max(20, degree * 10)
+        x_samples = np.linspace(-1.0, 1.0, n_samples)
+    x_samples = np.asarray(x_samples, dtype=np.float64)
+    if initial_phases is None:
+        initial_phases = RNG.normal(loc=0.0, scale=0.1, size=degree + 1)
+    target_values = np.asarray([target(float(x)) for x in x_samples], dtype=np.float64)
```
refraction-ray: for the dtype, also use set_dtype and unified tc.dtypestr and tc.rdtypestr to allow switch

examples/qsp.py
```diff
@@ -0,0 +1,208 @@
...
+    x_tensor = _real_tensor(x_samples)
+    target_tensor = _real_tensor(target_values)
+    use_gradient = tc.backend.name != "numpy"
+
+    def loss(phases: np.ndarray) -> object:
+        values = tc.backend.vmap(
+            lambda sample: _qsp_circuit_value(phases, sample), vectorized_argnums=0
+        )(x_tensor)
+        difference = values - target_tensor
+        return tc.backend.mean(difference * difference)
```
refraction-ray: sum on elementwise product is slower than tensordot

examples/qsvt.py
```diff
@@ -0,0 +1,186 @@
...
+
+This script contains a small QSVT solver and runs a basic identity-target
+demo when executed directly.
+"""
+
+from typing import Any, Callable, Optional
+
+import numpy as np
+
+from qsp import _fit_qsp_phases
```
refraction-ray: fit_qsp_phases should be a public method name convention

examples/qsvt.py
```diff
@@ -0,0 +1,186 @@
...
+    matrix_dag = tc.backend.adjoint(matrix_tensor)
+    identity = tc.backend.eye(matrix_tensor.shape[0], dtype=tc.dtypestr)
+    sqrt_right = tc.backend.sqrtmh(identity - matrix_tensor @ matrix_dag, psd=True)
+    sqrt_left = tc.backend.sqrtmh(identity - matrix_dag @ matrix_tensor, psd=True)
+    top = tc.backend.concat([matrix_tensor, sqrt_right], axis=1)
+    bottom = tc.backend.concat([sqrt_left, -matrix_dag], axis=1)
+    return tc.backend.concat([top, bottom], axis=0)
+
+
+def _leading_basis_inputs(system_dimension: int) -> object:
```
refraction-ray: dont define function for one liner if it is only called once

examples/qsvt.py
```diff
@@ -0,0 +1,186 @@
...
+    transformed_eigvals = np.asarray(
+        [target_function(float(np.real(eigval))) for eigval in eigvals],
+        dtype=np.complex128,
+    )
+    return eigvecs @ np.diag(transformed_eigvals) @ eigvecs.conj().T
+
+
+def _transform_qsp_phases_to_qsvt(phases: np.ndarray) -> np.ndarray:
+    """Convert QSP phases to the reflection-convention QSVT phases."""
+    qsp_phases = np.asarray(phases, dtype=np.float64)
```
refraction-ray: no numpy, backend agonistic

examples/qsvt.py
```diff
@@ -0,0 +1,186 @@
...
+    target = lambda x: x
+    sample_points = spectral_samples(matrix)
+    system_qubits = _system_qubit_count(matrix)
+    system_dimension = 2**system_qubits
+    total_qubits = system_qubits + 1
+
+    print("Target polynomial: f(x) = x")
+    print(f"Target matrix shape: {matrix.shape}")
+    print("Target matrix:")
+    with np.printoptions(precision=3, suppress=True):
```
refraction-ray: no need to print the matrix

examples/qsvt.py
```diff
@@ -0,0 +1,186 @@
...
+def spectral_samples(matrix: np.ndarray) -> np.ndarray:
+    spectrum = np.linalg.eigvalsh(np.asarray(matrix, dtype=np.complex128))
+    return np.asarray(np.real_if_close(spectrum), dtype=np.float64)
+
+
+def target_matrix(
+    matrix: np.ndarray,
+    target_function: Callable[[float], float],
+) -> np.ndarray:
+    eigvals, eigvecs = np.linalg.eigh(np.asarray(matrix, dtype=np.complex128))
```
refraction-ray: backend instead of numpy

examples/qsvt.py
```diff
@@ -0,0 +1,186 @@
...
+    return eigvecs @ np.diag(transformed_eigvals) @ eigvecs.conj().T
+
+
+def _transform_qsp_phases_to_qsvt(phases: np.ndarray) -> np.ndarray:
+    """Convert QSP phases to the reflection-convention QSVT phases."""
+    qsp_phases = np.asarray(phases, dtype=np.float64)
+    updates = np.empty_like(qsp_phases)
+    updates[0] = 3 * np.pi / 4 - (3 + len(qsp_phases) % 4) * np.pi / 2
+    updates[1:-1] = np.pi / 2
+    updates[-1] = -np.pi / 4
```
refraction-ray, line 66: comment on why this shift and convention mismatch

refraction-ray (changes requested):

examples/qsp.py
```diff
@@ -0,0 +1,213 @@
...
+    loss: Callable[[np.ndarray], object],
+    initial_parameters: np.ndarray,
+    maxiter: int,
+    use_gradient: bool,
+) -> object:
+    optimized_loss = tc.backend.jit(loss)
+    if tc.backend.name == "numpy":
+
+        def numpy_loss(phases: np.ndarray) -> float:
+            return float(tc.backend.numpy(optimized_loss(phases)))
```
refraction-ray: the float and backend.numpy are both unnecessary here?

examples/qsp.py
```diff
@@ -0,0 +1,213 @@
...
+            state,
+            parameters,
+            value=loss_value,
+            grad=gradients,
+            value_fn=bounded_loss,
+        )
+        parameters = optax.apply_updates(parameters, updates)
+        return parameters, state, loss_value, gradients
+
+    raw_parameters, optimizer_state, loss_value, gradients = update_step(
```
refraction-ray: why a separate first update step not in the for?

examples/qsp.py
```diff
@@ -0,0 +1,213 @@
...
+        raw_parameters, optimizer_state
+    )
+    for _ in range(1, maxiter):
+        raw_parameters, optimizer_state, loss_value, gradients = update_step(
+            raw_parameters, optimizer_state
+        )
+
+    bounded_parameters = np.pi * tc.backend.tanh(raw_parameters)
+    gradient_norm = float(tc.backend.numpy(tc.backend.norm(gradients)))
```
refraction-ray: why not keep gradient_norm backend format

examples/qsp.py
```diff
@@ -0,0 +1,213 @@
...
+) -> MinimizeResult:
+    if maxiter is None:
+        maxiter = 500
+    if x_samples is None:
+        n_samples = max(20, degree * 10)
+        x_samples = np.linspace(-1.0, 1.0, n_samples, dtype=tc.rdtypestr)
+    x_samples = np.asarray(x_samples, dtype=tc.rdtypestr)
+    if initial_phases is None:
+        initial_phases = RNG.normal(loc=0.0, scale=0.1, size=degree + 1)
+    initial_phases = np.asarray(initial_phases, dtype=tc.rdtypestr)
```
refraction-ray: this is unnecessary

examples/qsp.py
```diff
@@ -0,0 +1,213 @@
...
+    target_values = np.asarray(
```
refraction-ray: just backend.stack, instead first np then real tensor

examples/qsvt.py
```diff
@@ -0,0 +1,185 @@
...
+    basis_inputs = tc.backend.eye(2 * system_dimension, dtype=tc.dtypestr)[
+        :system_dimension
+    ]
+    columns = tc.backend.vmap(
+        lambda input_state: build_qsvt_circuit(
+            phases, unitary_matrix, total_qubits, inputs=input_state
+        ).state()[:system_dimension],
+        vectorized_argnums=0,
+    )(basis_inputs)
+    return tc.backend.transpose(columns)
```
refraction-ray: you can get the circuit matrix directly by c.matrix() instead of this function

examples/qsvt.py
```diff
@@ -0,0 +1,185 @@
...
+    matrix_tensor = _complex_tensor(matrix)
+    system_dimension = tc.backend.shape_tuple(matrix_tensor)[0]
+    total_qubits = _system_qubit_count(matrix_tensor) + 1
+    unitary_tensor = block_encode(matrix_tensor)
+    eigvals, eigvecs = tc.backend.eigh(matrix_tensor)
+    transformed_eigvals = _complex_tensor(
+        np.asarray(
+            [
+                target_function(float(value))
+                for value in np.asarray(tc.backend.numpy(tc.backend.real(eigvals)))
```
refraction-ray: np.asarray and backend numpy are all unessesary

refraction-ray (changes requested):
almost here

examples/qsp.py
```diff
@@ -73,31 +73,28 @@ def bounded_loss(raw_phases: object) -> object:
...
-
-    @tc.backend.jit
-    def update_step(
-        parameters: object,
-        state: object,
-    ) -> tuple[object, object, object, object]:
-        loss_value, gradients = value_and_grad(parameters)
-        updates, state = optimizer.update(
+    optimizer_update = tc.backend.jit(
+        lambda gradients, state, parameters, loss_value, value_fn: optimizer.update(
```
refraction-ray: dont over use lambda function, define a function for update

examples/qsp.py
refraction-ray: ?

examples/qsvt.py
```diff
@@ -118,20 +100,16 @@ def circuit_error(
...
-            dtype=tc.dtypestr,
-        )
-    )
+    transformed_eigvals = _complex_tensor(target_function(tc.backend.real(eigvals)))
     target_tensor = (
         eigvecs @ tc.backend.diagflat(transformed_eigvals) @ tc.backend.adjoint(eigvecs)
     )
     target_norm = tc.backend.norm(target_tensor)
-    approx_circ = circuit_block(phases, unitary_tensor, system_dimension, total_qubits)
+    with tc.runtime_contractor("plain"):
```
refraction-ray: why this, no need at all, and plain is a contractor that never should be used

examples/qsvt.py
```diff
@@ -0,0 +1,162 @@
...
+
+This script contains a small QSVT solver and runs a basic identity-target
+demo when executed directly.
+"""
+
+from typing import Any, Callable, Optional
+
+import numpy as np
+
+from qsp import fit_qsp_phases
```
refraction-ray, line 13: this should be at the bottom of import tensorcircuit

examples/qsvt.py
DeanTMaxim: Thank you for this insightful comment. We explicitly use the plain contractor here to ensure environment-wide robustness. In tensorcircuit, the .matrix() method internally introduces CopyNodes (hyperedges) to construct the identity transformation. Since the default greedy contractor requires the optional cotengra package to handle such hyperedge contractions, forcing the plain contractor allows the script to run "out-of-the-box" without external dependency issues, while maintaining sufficient performance for these matrix verifications.

refraction-ray: good point, but then the user should install cotengra, plain will never be used. example script never care about dependecies, unlike the codebase itself

DeanTMaxim: Thank you for the suggestion; we have removed the explicit plain contractor setting.

refraction-ray:
thanks for the contribution and the patience to continuously improve the scripts.

@all-contributors please @DeanTMaxim for example

refraction-ray:
@all-contributors please add @DeanTMaxim for example

PR #125, opened 2026-07-16
add unified contraction algebra interface with tropical and bf16 applications

DeanTMaxim:
## Summary

A generic, in-source `ContractionAlgebra` interface: run a tensor-network contraction over an arbitrary algebra (kernels + a boundary representation) via `tc.set_contractor(algebra=...)`, routed by `cons._algebraic_base_contraction`. Two reference applications in `applications/`:

- **tropical (max-plus)** — ground-state energy, configuration recovery, degeneracy counting (arXiv:2008.06888)
- **complex\<bfloat16\> pair-algebra** — complex multiply as 4 real bf16 matmuls

Standard contractions are unchanged.

## Motivation

tc-ng contractions normally run over the standard complex semiring. Some workloads need a different algebra — tropical (max-plus) for spin-glass ground states, or a low-precision storage (bf16 pair) for memory. Rather than special-case each in the contraction internals, this PR adds one generic entry point. The key choice is *where* a non-standard representation plugs in: the `_algebraic_base_contraction` boundary is the single place raw tensors are materialized outside the kernels, so a representation can transform storage once (leaves in, final out) without touching `tn.Node` or the topology. That keeps a non-standard axis (e.g. the bf16 pair axis) off `tn.Node`, which assumes axis==edge — avoiding that conflict by construction.

## Design

Two ABCs in `tensorcircuit/contraction_algebra/base.py`: `ContractionAlgebra` (tensordot/einsum kernels + optional hooks) carries a `Representation` (boundary `encode`/`decode` codec), **bundled** so a kernel can't be mis-paired with a codec. `_algebraic_base_contraction` runs `encode`→kernels→`decode` at the topology boundary (after `_extract_topology`, before `tn.Node`); three small in-source guards in `cons.py` route non-standard algebras and skip `preprocessing`. Counting's degeneracy rides an aux side-channel (`degeneracy()`).

## Breaking changes

**None for upstream `master`** — master has no `contraction_algebra` module; the diff is additive except ~11 internal lines in `cons.py`. Activate an algebra with `tc.set_contractor(algebra=...)`, or the `tropical()` / `counting_tropical()` / `bcomplex32()` context managers.

## Testing

~65 tests: standard==native keystone; tropical energy / counting (scalar + non-scalar) / tracking vs brute force; bf16 4M kernel units + end-to-end vs `complex64` + an axis==edge wall canary. Author-run gates green (`black`, `mypy --strict`, `pylint 10.00/10` on the new modules).

## Notes for review

- **bf16 is numpy-correctness only** this PR; GPU (jax/torch/cupy) bf16 kernels + benchmark are a follow-up (the codec is already backend-agnostic). numpy `einsum` rejects bf16, so `_pair_einsum` upcasts bf16→float32 per call (structure/quantization unchanged).
- **`ml_dtypes` is currently undeclared** (lazily imported by the bf16 path; not in `pyproject.toml`) — maintainer decision whether to add it (e.g. an optional extra).

## Out of scope (follow-up PRs)

GPU bf16 kernel + benchmark (L3); `set_dtype("bcomplex32")` pipeline (L4); non-scalar tracking `recover_configuration` (orthogonal, separately gated).


examples/tropical_ising.py
```diff
@@ -0,0 +1,71 @@
...
+    return nodes
+
+
+def main():
+    n = 5
+    edges = [(0, 1), (1, 2), (2, 3), (3, 4), (4, 0)]
+    J = [1.0, -1.0, 1.0, 1.0, -1.0]
+    h = [0.5, -0.3, 0.2, 0.0, 0.4]
+    nodes = build_tn(n, edges, J, h)
+    activate()
```
refraction-ray: why this is not integrated into the contextmanager: the activate and finally teardown part

applications/tropical_algebra.py
```diff
@@ -0,0 +1,716 @@
...
+from tensorcircuit.contraction_algebra import (
+    ContractionAlgebra,
+    activate,
+    set_contraction_algebra,
+    get_contraction_algebra,
+)
+
+Tensor = Any
+Backend = Any
+_EPS = 1e-9
```
refraction-ray, line 45: is this sufficient at FP64?

tests/_tropical_test_utils.py
```diff
@@ -0,0 +1,101 @@
+"""Shared helpers for the tropical Ising tensor-network tests.
```
refraction-ray: dont use so many small test files, better a tropical test file together

CHANGELOG.md
```diff
@@ -38,6 +38,8 @@
 
 - Add pure-Python pytree utilities for the abstract backend.
 
+- Add `contraction_algebra` package: generic runtime monkey-patch for swapping contraction primitives (tensordot/einsum), with a tropical (max-plus) reference application (ground-state energy / configuration recovery / degeneracy counting, arXiv:2008.06888).
```
refraction-ray: it is not a package it is a submodule

pyproject.toml
```diff
@@ -52,6 +52,7 @@ filterwarnings = [
     "ignore::DeprecationWarning",
     "ignore:Explicitly requested dtype*:UserWarning",
 ]
+pythonpath = ["."]
```
refraction-ray: don't add your local development practice, the tensorcircuit-ng package can be installed in develop mode instead of pointer to the project path

examples/tropical_ising.py
```diff
@@ -0,0 +1,71 @@
+"""Tropical (max-plus) Ising ground-state energy via TensorCircuit-NG contraction.
+
+Run: PYTHONPATH=<repo> python examples/tropical_ising.py
```
refraction-ray: no the env, it is your personal preference not a necessary env variable

tensorcircuit/contraction_algebra.py
```diff
@@ -0,0 +1,46 @@
+"""ContractionAlgebra abstraction: a semiring descriptor consulted during contraction."""
+
+from abc import ABC, abstractmethod
+from typing import Any
+
+Tensor = Any
+Backend = Any
+
+
+class ContractionAlgebra(ABC):
```
refraction-ray, line 53: the interface design and impl require careful consideration, especially concerning that in low precison model, the Node content itself should be patched at two real tensors

tensorcircuit/contraction_algebra/injection.py
```diff
@@ -0,0 +1,160 @@
...
+def runtime_contraction_algebra() -> StandardAlgebra:
+    return cast(StandardAlgebra, _state["algebra"])
+
+
+def set_contraction_algebra(alg: Any) -> None:
+    _state["algebra"] = alg
+
+
+def get_contraction_algebra() -> StandardAlgebra:
+    return cast(StandardAlgebra, _state["algebra"])
```
refraction-ray: what does this cast for? dont add function and maitain burden only for mypy pass, just use type ignore comment or Any type instead

refraction-ray (changes requested):
I am just quickly going through and raise some more evident catches, the most important on is on the interface deisgn, you need to consider the node patch case together to design the interface in one run, instead of further changing the interface later

examples/tropical_ising.py
DeanTMaxim: build_tn is problem-specific Ising encoding that doesn't depend on the contraction algebra.

applications/tropical_algebra.py
DeanTMaxim: 1e-9 gives 5 orders of magnitude headroom above FP64 accumulated rounding error for the demonstrated use cases (N≤8, J1), and the mask only matters for true ties — non-tie positions differ by O(1) and are unaffected.

applications/bcomplex32_algebra.py
```diff
@@ -0,0 +1,194 @@
...
+    re = be.cast(be.real(t), bf)
+    im = be.cast(be.imag(t), bf)
+    return be.stack([re, im], axis=-1)
+
+
+def _pair_to_complex(be: Backend, pair: Tensor) -> Tensor:
+    """pair of bf16 -> complex64 tensor (recombine; no copy risk via cast)."""
+    re = be.cast(pair[..., 0], "float32")
+    im = be.cast(pair[..., 1], "float32")
+    return be.cast(re + 1j * im, "complex64")
```
refraction-ray: this is subtle, the target should follow tc.cons.dtype?

tensorcircuit/contraction_algebra/base.py
```diff
@@ -0,0 +1,95 @@
...
+    cotengra decomposes any contraction into pairwise steps; each is a
+    ``tensordot`` (ordinary pair) or ``einsum`` (hyperedge / copy-node). An
+    algebra supplies both. The kernel must be closed over the layout its
+    ``representation`` produces.
+    """
+
+    name: str = "abstract"
+    representation: Representation = IdentityRepresentation()
+    # When operands carry a trailing non-physical storage axis (e.g. the
+    # complex<bf16> pair), cotengra's tensordot mode post-transposes results via
```
refraction-ray: but you can customize the tensordot, can you avoid this?

tensorcircuit/contraction_algebra/__init__.py
```diff
@@ -0,0 +1,27 @@
+"""ContractionAlgebra: a generic interface for swapping contraction primitives +
```
refraction-ray: whether a file of contraction_algebra is sufficient? no need for a submodule?

tensorcircuit/cons.py
```diff
@@ -138,6 +139,43 @@ def set_tensornetwork_backend(
 
 set_tensornetwork_backend()
 
+# --- ContractionAlgebra state (mirrors dtypestr/set_dtype pattern) ---
+from .contraction_algebra import StandardAlgebra as _StandardAlgebra
```
refraction-ray: no middle import

tensorcircuit/cons.py
```diff
@@ -138,6 +139,43 @@ def set_tensornetwork_backend(
...
+from .contraction_algebra.base import (
+    ContractionAlgebra as _ContractionAlgebra,
+)
+
+_contraction_algebra: _ContractionAlgebra = _StandardAlgebra()
```
refraction-ray: this will slowdown the default running as a more layer abstraction. I think the default contraction_algebra is better be None. and if it is None, the whole logic in contractor in terms of algebra is just skipped

refraction-ray: by this, the original code has a clear branch with no diff, which better ensures minimal change, backward compatibility and reviewer friendliness

applications/tropical_algebra.py
```diff
@@ -0,0 +1,811 @@
...
+configuration / degeneracy) from Liu, Wang, Zhang PRL 126, 090506 (2021)
+(arXiv:2008.06888). Contract under a tropical algebra via::
+
+    from applications.tropical_algebra import tropical
+    with tropical():
+        ...  # max-plus contraction
+
+Sections (consolidated from the original package modules):
+  1. max-plus primitives + MaxPlusAlgebra              (was tensorcircuit/contraction_algebra/tropical.py)
+  2. counting (energy, degeneracy) + split_energy_count (was .../counting.py)
```
refraction-ray, line 25: was is unnecessary

tensorcircuit/cons.py
```diff
@@ -138,6 +139,43 @@ def set_tensornetwork_backend(
...
+    global _contraction_algebra
+    _contraction_algebra = alg
+
+
+# Aux output side-channel for multi-output (non-standard) algebras. Cleared and
+# refilled on each non-standard contraction by ``_stash_aux_outputs``; the
+# primary tensor is returned through the normal contraction return value, while
+# aux (e.g. counting degeneracy, sharing the primary's physical axis order) is
+# stashed here for the caller to read via ``_aux_outputs``.
+_aux_outputs_store: Dict[str, Any] = {}
```
refraction-ray: is this necessarily a global variable, what if it is the arguments for the below functions?

tensorcircuit/cons.py
```diff
@@ -735,48 +741,74 @@ def _wrap_omeco_optimizer(optimizer: Any) -> Any:
...
+    size_dict: Any,
+    algorithm: Any,
+    ns: bool,
+    alg: Any,
+    kbe: Any,
+    be: Any,
+    strip_exponent: bool,
+    ctg: Any,
+) -> Any:
+    """Run the (possibly single-tensor) contraction, returning (final, exponent)."""
```
refraction-ray: no docstrings for these mysterious variables

refraction-ray: specifically the name convention with both algorithm and alg is bad

applications/tropical_algebra.py
```diff
@@ -0,0 +1,811 @@
...
+
+    _finalize_from_leaves(tree, opt_pos, assignment)
+    return assignment
+
+
+# ===== Section 4: context managers =====
+
+
+@contextlib.contextmanager
+def tropical(track: bool = False) -> Iterator[None]:
```
refraction-ray, line 748: just directly reuse the contractor context manager in cons, instead of defining each one separately

applications/bcomplex32_algebra.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+def _pair_tensordot(be: Backend, a: Tensor, b: Tensor, axes: Any) -> Tensor:
+    """Complex tensordot = 4 real bf16 tensordots (4M). Uses be.tensordot (never patched)."""
+    ar, ai = a[..., 0], a[..., 1]
+    br, bi = b[..., 0], b[..., 1]
+    cr = be.tensordot(ar, br, axes) - be.tensordot(ai, bi, axes)
+    ci = be.tensordot(ar, bi, axes) + be.tensordot(ai, br, axes)
+    return be.stack([cr, ci], axis=-1)
```
refraction-ray: Storing real and image in a single [..., 2] tensor forces the code to constantly slice (pair[..., 0] and pair[..., 1]), mask, and restack using _stack_last. This is computationally wasteful and hard to read.

Refactor: Keep the energy and count as a standard Python Tuple[Tensor, Tensor] throughout the computation. Only the framework's boundary layers (encode and decode) need to know they are paired. Passing tuples avoids backend-specific stacking bottlenecks and simplifies the math logic. This also apply to tropical_algebra case. Keep it as pair instead on stack them into one tensor

applications/tropical_algebra.py
```diff
@@ -0,0 +1,811 @@
...
+
+
+class MaxPlusAlgebra(ContractionAlgebra):
+    """Tropical (max, +) semiring: addition -> max, multiplication -> +."""
+
+    name = "maxplus"
+    # representation defaults to IdentityRepresentation (real tensors, no codec):
+    # leaves enter/exit the contraction unchanged; only the kernels differ.
+
+    def tensordot(self, be: Backend, a: Tensor, b: Tensor, axes: Any) -> Tensor:
```
refraction-ray, line 156: deserve a try later, whe decorated with jax.jit, whether the actual kernel can autoamtically compile to a space memory balance version with chucks automatically. the current broadcast impl is memory intensive

refraction-ray: have already tried，the answer is no, jax.jit is not that smart

requirements/requirements-dev.txt
```diff
@@ -3,7 +3,7 @@ pytest==7.4.4
 pytest-cov
 pytest-benchmark
 pytest-xdist
-black[jupyter]
+black[jupyter]==26.5.1
```
refraction-ray: no need to fix this, we always keep black to the latest version

tests/_tropical_test_utils.py
DeanTMaxim: Fixed.

CHANGELOG.md
DeanTMaxim: Fixed.

tensorcircuit/contraction_algebra/base.py
DeanTMaxim: Good catch — prefer_einsum is indeed an implementation detail of how cotengra post-transposes after tensordot (the permutation doesn't account for the non-physical pair axis, and we can't intercept that from inside tensordot), so I'll extract it into a get_contractor_kwargs() method to keep the ABC clean while preserving the necessary escape hatch.

docs/2026-07-19-contractor-kwargs.md
```diff
@@ -0,0 +1,173 @@
+# Replace `prefer_einsum` with `get_contractor_kwargs()` — Implementation Plan
```
refraction-ray: dont push your local spec files

tensorcircuit/cons.py
DeanTMaxim: _aux_outputs_store removed — aux is now stored on the CountingRepresentation that produces it, so cons.py has zero algebra-specific globals.

tensorcircuit/cons.py
```diff
@@ -21,6 +22,10 @@
 from .backends.numpy_backend import NumpyBackend
 from .backends import get_backend
 from .simplify import _multi_remove
+from .contraction_algebra import (
+    ContractionAlgebra as _ContractionAlgebra,
```
refraction-ray, line 26: why as, and there is also lower case `_contraction_algebra` variable in this file, dangerously similar and confusing

tensorcircuit/cons.py
```diff
@@ -138,6 +143,19 @@ def set_tensornetwork_backend(
...
+# --- ContractionAlgebra state (mirrors dtypestr/set_dtype pattern) ---
+
+_contraction_algebra: Optional[_ContractionAlgebra] = None
+
+
+def get_contraction_algebra() -> Optional[_ContractionAlgebra]:
+    return _contraction_algebra
+
+
+def set_contraction_algebra(alg: Optional[_ContractionAlgebra]) -> None:
```
refraction-ray, line 155: no support for default arg, say `set_contraction_algebra()` with no arg should recover the default behavior

tensorcircuit/cons.py
```diff
@@ -735,48 +721,82 @@ def _wrap_omeco_optimizer(optimizer: Any) -> Any:
...
-    final_node = tn.Node(final_raw_tensor, backend=be)
+            final, exponent = tree.contract(raw_tensors, strip_exponent=True)
+    return final, exponent
+
+
+def _decode(ns: bool, rep: Any, be: Any, final: Any, output_set: Any) -> Any:
+    """Decode the contraction output under a non-standard algebra.
+
+    ``rep.decode`` is responsible for stashing any aux internally."""
+    if not ns:
```
refraction-ray, line 786: no need for ns, if not ns, in the parent function there should be no call on _decode

tensorcircuit/cons.py
```diff
@@ -786,12 +806,72 @@ def _algebraic_base_contraction(
...
+    import cotengra as ctg
+
+    raw_tensors, input_sets, output_set, size_dict = _extract_topology(nodes)
+
+    alg = get_contraction_algebra()
+
+    if alg is not None:
+        rep = alg.representation
+    else:
+        rep = IdentityRepresentation()  # no-op; _decode skips it when ns=False
```
refraction-ray, line 829: better no need to set?

tensorcircuit/cons.py
```diff
@@ -786,12 +806,72 @@ def _algebraic_base_contraction(
...
+        raw_tensors = rep.encode(backend, raw_tensors)
+
+    final, exponent = _run_contraction(
+        raw_tensors,
+        input_sets,
+        output_set,
+        size_dict,
+        algorithm,
+        alg is not None,
+        algebra=alg,
```
refraction-ray, line 845: merge the two variable? as algebra=None means ns=False

tensorcircuit/cons.py
```diff
@@ -786,12 +806,72 @@ def _algebraic_base_contraction(
...
+        be=backend,
+        strip_exponent=kws.get("strip_exponent", False),
+        ctg=ctg,
+    )
+
+    final = _decode(alg is not None, rep, backend, final, output_set)
```
refraction-ray, line 851: if outside the _decode call

tensorcircuit/cons.py
```diff
@@ -1186,6 +1267,11 @@ def set_contractor(
     To set runtime contractor of the tensornetwork for a better contraction path.
     For more information on the usage of contractor, please refer to independent tutorial.
 
+    To change the contraction algebra, use ``cons.set_contraction_algebra(alg)``
+    separately (the algebra is orthogonal to the contractor configuration).  The
+    ``tropical()`` / ``bcomplex32()`` / ``counting_tropical()`` context managers
```
refraction-ray, line 1272: no, dont define these sub context managers, just use something like `set_contraction_algebra(TropicalAlgebra)`

tensorcircuit/contraction_algebra.py
```diff
@@ -0,0 +1,154 @@
...
+    name = "identity"
+
+    def encode(self, be: Backend, tensors: List[Tensor]) -> List[Tensor]:
+        return tensors
+
+    def decode(self, be: Backend, tensor: Tensor) -> Tuple[Tensor, Dict[str, Tensor]]:
+        return tensor, {}
+
+
+class ContractionAlgebra(ABC):
```
refraction-ray, line 53: is there elegant way to merge Representation class and Algebra class?

refraction-ray: is it okay just define encode and decode in Algebra class?

tensorcircuit/contraction_algebra.py
```diff
@@ -0,0 +1,154 @@
...
+    name = "standard"
+
+    def tensordot(self, be: Backend, a: Tensor, b: Tensor, axes: Any) -> Tensor:
+        return be.tensordot(a, b, axes)
+
+    def einsum(self, be: Backend, eq: str, *operands: Tensor) -> Tensor:
+        return be.einsum(eq, *operands)
+
+
+class PairTensor:
```
refraction-ray, line 108: why bother the class, just bare tuple suffices?

applications/bcomplex32_algebra.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def tensordot(self, be: Backend, a: Tensor, b: Tensor, axes: Any) -> Tensor:
+        return _pair_tensordot(be, a, b, axes)
+
+    def einsum(self, be: Backend, eq: str, *operands: Tensor) -> Tensor:
+        return _pair_einsum(be, eq, *operands)
+
+
+@contextlib.contextmanager
+def bcomplex32() -> Iterator[None]:
```
refraction-ray, line 171: delete this

refraction-ray (changes requested):

tests/test_bcomplex32_algebra.py
```diff
@@ -0,0 +1,129 @@
...
+    )  # reduce over b (sum in complex -> here identity semantics)
+    ref = np.einsum("ab->a", a)
+    np.testing.assert_allclose(
+        np.asarray(_pair_to_complex(be, out)), ref, rtol=2e-2, atol=5e-3
+    )
+
+
+# --- Task 12: end-to-end through real tc.Circuit + wall-avoidance canary ---
+
+import tensorcircuit as tc
```
refraction-ray, line 50: middle import is not allowed

tests/test_bcomplex32_algebra.py
```diff
@@ -0,0 +1,129 @@
...
+from applications.bcomplex32_algebra import bcomplex32
```
refraction-ray, line 51: how can you import something from application instead of tensoricrcuit.application?

tests/test_bcomplex32_algebra.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+def test_bf16_end_to_end_matches_complex64():
```
refraction-ray, line 54: set jax, tf and numpy backend with fixtures?

tests/test_bcomplex32_algebra.py
```diff
@@ -0,0 +1,129 @@
...
+
+    with bcomplex32():
+        c = tc.Circuit(4)
+        for i in range(4):
+            c.H(i)
+        for i in range(3):
+            c.cnot(i, i + 1)
+        st = np.asarray(c.state())
+    assert st.shape == (16,)  # ran cleanly, no axis==edge crash
+    import tensorcircuit.cons as cons
```
refraction-ray, line 81: no middle import, too ugly

tests/test_contraction_algebra.py
```diff
@@ -0,0 +1,470 @@
+import numpy as np
+import tensorcircuit as tc
+import tensorcircuit.cons as cons
```
refraction-ray, line 3: better just use tc.cons

tests/test_contraction_algebra.py
```diff
@@ -0,0 +1,470 @@
...
+    ref = np.array(be.einsum("ab,bc->ac", a, b))
+    np.testing.assert_allclose(got, ref)
+
+
+def test_standard_name():
+    assert StandardAlgebra().name == "standard"
+
+
+def test_public_api_surface():
+    from tensorcircuit import contraction_algebra as tca
```
refraction-ray, line 41: no middle and repeated import

tests/test_contraction_algebra.py
```diff
@@ -0,0 +1,470 @@
...
+
+    # After Task 14 the package exports only the 4 base names; activation lives
```
refraction-ray, line 43: dont leave think loudly type comments

tests/test_contraction_algebra.py
```diff
@@ -0,0 +1,470 @@
...
+    # in-source via cons.set_contraction_algebra.
+    for name in [
+        "ContractionAlgebra",
+        "StandardAlgebra",
+        "Representation",
+        "IdentityRepresentation",
+    ]:
+        assert hasattr(tca, name), name
+    # The old monkey-patch API names are intentionally gone:
```
refraction-ray, line 52: why you test something are changed to ensure they are not existing anymore, weird...

tests/test_contraction_algebra.py
```diff
@@ -0,0 +1,470 @@
...
+    assert alg.on_contraction_start(["dummy_nodes"]) is None
+    assert alg.on_contractor_ready(["dummy_tree"]) is None
+
+
+from tensorcircuit.contraction_algebra import (
+    ContractionAlgebra,
+    StandardAlgebra,
+    Representation,
+    IdentityRepresentation,
+)
```
refraction-ray, line 76: middle import again

tests/test_contraction_algebra.py
```diff
@@ -0,0 +1,470 @@
...
+        assert log.index("ready") < log.index("decode")
+    finally:
+        cons.set_contraction_algebra(prev)
+
+
+def test_decode_wrong_rank_raises_loudly():
+    import numpy as np
+    import tensornetwork as tn
+    import opt_einsum
+    import pytest
```
refraction-ray, line 144: ...

PR #127, opened 2026-07-21
add TianYan cloud provider

QuiXamii:
- new tensorcircuit/cloud/tianyan.py provider integrated into cloud apis: device listing, standardized properties, task submission (single/batch), resubmission, and result query via the cqlib SDK
- automatic topology-aware qubit mapping for real hardware submissions
- bit ordering convention verified on tianyan_sw simulator; basic chain verified on tianyan176 hardware (Bell 00+11 ratio 0.86 with 50 shots)
- offline unit tests with mocked platform plus env-gated online simulator and hardware tests
- docs: API reference, EN/CN tutorials and reference pages, demo example


docs/source/tutorials/tianyan_cloud_reference.md
```diff
@@ -0,0 +1,226 @@
...
+)
+```
+
+## Compilation and topology mapping
+
+When a TensorCircuit circuit is submitted to real hardware, TianYan automatically:
+
+1. Queries the device topology.
+2. Finds physical qubits that satisfy the circuit interactions.
+3. Maps logical qubits to physical qubits.
```
refraction-ray: how this is done? automatically under tianyan sdk, tianyan server or we handle this?

docs/source/tutorials/tianyan_cloud_reference.md
```diff
@@ -0,0 +1,226 @@
...
+4. Keep hardware shots small to avoid unnecessary resource consumption.
+5. The example and notebook skip hardware by default. Set `TIANYAN_RUN_HARDWARE=1` explicitly to enable it.
+6. Measurements are submitted as terminal measurements in their recorded order. Mid-circuit measurement semantics are not preserved; do not submit circuits that depend on them.
+
+## Troubleshooting
+
+### `cqlib` is not installed
+
+```bash
+pip install "cqlib>=1.3.10,<1.4"
```
refraction-ray: why < 1.4

tensorcircuit/cloud/tianyan.py
```diff
@@ -0,0 +1,844 @@
...
+    extra_qir = circuit._extra_qir
+    qubits = set()
+    for inst in qir:
+        for idx in inst.get("index", ()):
+            qubits.add(idx)
+    for inst in extra_qir:
+        if inst.get("name") == "measure":
+            idx = inst.get("index", [None])[0]
+            if idx is not None:
+                qubits.add(idx)
```
refraction-ray: why not just range(nqubits)?

tensorcircuit/cloud/tianyan.py
```diff
@@ -0,0 +1,844 @@
...
+            used_physical.add(physical)
+            if search(position + 1):
+                return True
+            used_physical.remove(physical)
+            del mapping[logical]
+        return False
+
+    if search(0):
+        return mapping
+    raise ValueError("Cannot map all circuit interactions to connected physical qubits")
```
refraction-ray: we better dont mannually maintain logic qubit mapping here, the user should use compiler module for this, if the submited circuit is not compatible with the backend device topology, just raise ValueError

tests/test_tianyan.py
```diff
@@ -0,0 +1,676 @@
...
+
+SIMULATOR = "tianyan_sw"  # full-amplitude simulator, no topology constraints
+REAL_DEVICE = "tianyan176"  # superconducting hardware with topology constraints
+
+# keep shots low on real hardware to save quota
+TEST_SHOTS_SIM = 1000
+TEST_SHOTS_REAL = 100
+
+
+@unittest.skipUnless(HAS_CQLIB, "cqlib is required for TianYan conversion tests")
```
refraction-ray: use pytest skipif

tests/test_tianyan.py
```diff
@@ -0,0 +1,676 @@
...
+class TestQasmConversion(unittest.TestCase):
```
refraction-ray: better not use unittest, use pytest syntax

tests/test_tianyan.py
```diff
@@ -0,0 +1,676 @@
...
+        c = tc.Circuit(1)
+        c.h(0)
+        c.measure_instruction(0)
+
+        t = apis.submit_task(circuit=c, device=device, shots=50)
+        counts = t.results(blocked=True)
+        self.assertIsCounts(counts)
+
+
+if __name__ == "__main__":
```
refraction-ray: no need fot this

tests/test_tianyan.py
```diff
@@ -0,0 +1,676 @@
...
+        """A single-qubit circuit needs no topology mapping; use the simulator."""
+        device = apis.get_device(f"tianyan::{SIMULATOR}")
+
+        c = tc.Circuit(1)
+        c.h(0)
+        c.measure_instruction(0)
+
+        t = apis.submit_task(circuit=c, device=device, shots=50)
+        counts = t.results(blocked=True)
+        self.assertIsCounts(counts)
```
refraction-ray: use pytest syntax instead of unitest one

examples/tianyan_cloud_demo.py
```diff
@@ -0,0 +1,86 @@
...
+import os
+
+import tensorcircuit as tc
+
+SIMULATOR = "tianyan_sw"
+REAL_DEVICE = "tianyan176"
+
+
+def _get_login_key() -> str:
+    login_key = os.getenv("LOGIN_KEY")
```
refraction-ray: the env variable name better begin with TIANYAN, otherwise what if we also provide quafu key

refraction-ray (changes requested):

refraction-ray:
ccx/toffoli dispatch broken (tianyan.py:327) — `_tc_qir_to_qcis` dispatches on name == "ccx", but tc's QIR emits "toffoli" (ccx is just an alias, abstractcircuit.py:72). Any circuit using the gate advertised in native_gates raises Unsupported gate. fredkin/cswap unhandled too. I verified this directly against the source. Also missing: cy, iswap.

examples/tianyan_cloud_demo.py
refraction-ray: better reuse the env system I have rewritten in apis.py for set and get token

tensorcircuit/cloud/tianyan.py
```diff
@@ -0,0 +1,745 @@
...
+    Measurements are always emitted as terminal measurements in record
+    order; mid-circuit measurement semantics are not preserved.
+    """
+    qir = circuit.to_qir()
+    extra_qir = circuit._extra_qir
+    measure_instructions = [inst for inst in extra_qir if inst.get("name") == "measure"]
+    measure_instructions.sort(key=lambda inst: inst.get("record_index", 0))
+
+    _assert_cqlib()
+    from cqlib.circuits import Circuit as CqlibCircuit
```
refraction-ray: why this import within the function?

docs/source/tutorials/tianyan_cloud_reference.md
QuiXamii: It was previously handled inside tianyan.py (this provider). That logic has been removed in fecc104a: mapping is now the user's responsibility (e.g. via tensorcircuit.compiler), and the provider only validates topology compatibility before submission, raising ValueError for incompatible circuits. The docs have been updated accordingly.

docs/source/tutorials/tianyan_cloud_reference.md
QuiXamii: The provider hasn't been tested against cqlib 1.4 (it's not even released yet, latest is 1.3.11), so we cap the version to make sure users don't hit untested breakage.

tensorcircuit/cloud/tianyan.py
QuiXamii: This helper was only used by the provider-side mapping logic. Since that logic has been removed per the comment below, this function no longer exists.

docs/source/api/cloud.rst
```diff
@@ -6,6 +6,7 @@ tensorcircuit.cloud
     cloud/config.rst
     cloud/local.rst
     cloud/quafu_provider.rst
+    cloud/tianyan.rst
```
refraction-ray: this is added by hand instead of generated_rst.py? why ti is before te

docs/source/tutorials/tianyan_cloud_reference_cn.md
```diff
@@ -0,0 +1,224 @@
+# 天衍（TianYan）量子云平台接入说明
```
refraction-ray: just delete the two md files in tutorials, the content if necessary should move to the ipynb or the docstring in the source code

examples/tianyan_cloud_demo.py
```diff
@@ -0,0 +1,92 @@
...
+import os
+
+import tensorcircuit as tc
+
+SIMULATOR = "tianyan_sw"
+REAL_DEVICE = "tianyan176"
+
+
+def _get_login_key() -> str:
+    login_key = os.getenv("TIANYAN_LOGIN_KEY")
```
refraction-ray: the current name covention for the new token impl is `TC_TOKEN_`, please refer the new logic in apis.py for token setting and getting

examples/tianyan_cloud_demo.py
```diff
@@ -0,0 +1,92 @@
...
+    circuit.h(0)
+    circuit.cx(0, 1)
+    circuit.measure_instruction(0, 1)
+    return circuit
+
+
+def main() -> None:
+    """Run simulator workflows and optionally submit to real hardware."""
+    login_key = _get_login_key()
+    run_hardware = os.getenv("TIANYAN_RUN_HARDWARE") == "1"
```
refraction-ray: just run on hardware or this is more like a arg switch instead of env var

tensorcircuit/cloud/tianyan.py
QuiXamii: Done in fecc104a. The provider no longer maintains logical-to-physical qubit mapping.

examples/tianyan_cloud_demo.py
```diff
@@ -0,0 +1,92 @@
...
+        return
+
+    real_device = tc.cloud.apis.get_device(f"tianyan::{REAL_DEVICE}")
+    # real hardware requires a circuit that already respects the device topology;
+    # build a Bell circuit directly on a connected physical pair
+    q1, q2 = sorted(real_device.topology()[0])
+    hardware_circuit = tc.Circuit(q2 + 1)
+    hardware_circuit.h(q1)
+    hardware_circuit.cx(q1, q2)
+    hardware_circuit.measure_instruction(q1, q2)
```
refraction-ray: why we build a circuit again instead of submiting the same circuit to the hardware?

tests/test_tianyan.py
QuiXamii: Done in 3a90b949. The whole test file has been rewritten in pytest style.

tests/test_tianyan.py
QuiXamii: Done in 3a90b949. The whole test file has been rewritten in pytest style.

tests/test_tianyan.py
QuiXamii: Done in 3a90b949.

tests/test_tianyan.py
```diff
@@ -0,0 +1,644 @@
...
+
+LOGIN_KEY = os.getenv("TIANYAN_LOGIN_KEY")
+HAS_CQLIB = tianyan.TianYanPlatform is not None
+CLOUD_TESTS_ENABLED = (
+    HAS_CQLIB and bool(LOGIN_KEY) and os.getenv("TC_CLOUD_TEST") == "1"
+)
+HARDWARE_TESTS_ENABLED = (
+    CLOUD_TESTS_ENABLED and os.getenv("TC_CLOUD_HARDWARE_TEST") == "1"
+)
+FAKE_QUANTUM_LANGUAGE = SimpleNamespace(QCIS="QCIS")
```
refraction-ray, line 31: what is this for?

tests/test_tianyan.py
QuiXamii: Done in 3a90b949, same rewrite as above — all self.assertXxx calls are now plain assert statements.

examples/tianyan_cloud_demo.py
QuiXamii: Done in aef719a0. Switched to the shared env token system: the login key is now read from TC_TOKEN_TIANYAN automatically, and the explicit set_token calls were removed from the example, tests, docs and notebooks.

tensorcircuit/cloud/tianyan.py
```diff
@@ -0,0 +1,745 @@
...
+
+    # Ensure exp_name has a default
+    if exp_name is None:
+        exp_name = f"tc_exp_{datetime.now().strftime('%Y%m%d_%H%M%S')}"
+
+    # Submit
+    if is_sequence(source):
+        sources = list(source)
+        num_shots = _normalize_shots(shots, len(sources))
+        query_ids = pf.submit_experiment(
```
refraction-ray, line 669: what is the difference between submit experiment and submit job, submit experiment here is like submit a job group?

tensorcircuit/cloud/tianyan.py
```diff
@@ -0,0 +1,745 @@
...
+                        elif "gate error" in param_name:
+                            bits[q_idx]["SingleQubitErrRate"] = data["param_list"][idx]
+                        elif "Readout Error" in param_name:
+                            bits[q_idx]["ReadoutError"] = data["param_list"][idx]
+                        elif "|0> readout fidelity" in param_name:
+                            bits[q_idx]["ReadoutF0Err"] = 1.0 - data["param_list"][idx]
+                        elif "|1> readout fidelity" in param_name:
+                            bits[q_idx]["ReadoutF1Err"] = 1.0 - data["param_list"][idx]
+                        elif "f01" in param_name:
+                            bits[q_idx]["Freqency"] = data["param_list"][idx]
```
refraction-ray: frequency is a typo, though the typo may also exist in other previous cloud backends ==

tensorcircuit/cloud/tianyan.py
```diff
@@ -0,0 +1,745 @@
...
+    if task.device is None:
+        raise ValueError("Task must have an associated device for resubmission.")
+    pf = _get_platform(token, machine_name=task.device.name)
+    try:
+        # cqlib re-execute task
+        new_query_id = pf.re_execute_task(task.id_)
+        if isinstance(new_query_id, list):
+            new_query_id = new_query_id[0]
+        return Task(id_=new_query_id, device=task.device)
+    except Exception as e:
```
refraction-ray: why this except?

refraction-ray: we better use specific exception class instead of general Exception

refraction-ray:
In `_parse_result`, `measure_order = result_status[0]` is stored in `details["measure_order"]` but the bitstring is built by direct `"".join(str(b) for b in shot)`. If the platform ever returns shots ordered by `measure_order` rather than by qubit index, results silently permute.

refraction-ray:
`tianyan.py` uses `dict[int, set[int]]` and `tuple[dict[int, set[int]], set[int]]` while the rest of `cloud/` (and the same file's other signatures) uses `Dict`/`Tuple` from `typing`. inconsistent within the module set.

tests/test_tianyan.py
QuiXamii: It's a stand-in for cqlib's QuantumLanguage enum, used to patch it in offline tests where cqlib may not be installed. Added a clarifying comment in e60eeb48.

tensorcircuit/cloud/tianyan.py
QuiXamii: Fixed in e60eeb48.

docs/source/api/cloud.rst
QuiXamii: Entries now follow the alphabetical order.

docs/source/tutorials/tianyan_cloud_reference_cn.md
QuiXamii: Done in e60eeb48.

examples/tianyan_cloud_demo.py
QuiXamii: Done in aef719a0.

examples/tianyan_cloud_demo.py
QuiXamii: Done in e60eeb48. The example now uses an argparse --hardware flag instead of the env var. The notebooks keep the env var since they don't take CLI arguments.

examples/tianyan_cloud_demo.py
QuiXamii: Because the provider no longer remaps qubits: the simulator Bell circuit acts on logical qubits (0, 1), which are not necessarily coupled on the real device, and would now fail topology validation before submission.

refraction-ray: then better do a mapping manually on the old circuit, using `tc.Circuit.initial_mapping`

tensorcircuit/cloud/tianyan.py
QuiXamii: Added a clarifying comment in 3659f8a9. Submit_experiment submits the circuits as one experiment (a job group under lab_name) and returns one query id per circuit, while submit_job is the single-circuit endpoint.

tensorcircuit/cloud/tianyan.py
QuiXamii: Fixed in 1e784d94. The tianyan provider now uses the correct Frequency spelling. Note this makes the two providers' properties formats diverge on this key (tencent still emits Freqency). Shall we fix the typo in tencent.py as well?

tensorcircuit/cloud/tianyan.py
QuiXamii: The wrap mirrors tencent's resubmit_task, which also surfaces failures as ValueError with the task id. Switched to the specific CqlibError base class (covers CqlibRequestError, CqlibInputParaError, etc.) with exception chaining in 3659f8a9.

QuiXamii:
> In `_parse_result`, `measure_order = result_status[0]` is stored in `details["measure_order"]` but the bitstring is built by direct `"".join(str(b) for b in shot)`. If the platform ever returns shots ordered by `measure_order` rather than by qubit index, results silently permute.

Fixed in 0efedff6: bits are now re-emitted in ascending qubit-index order using measure_order (shot[i] is the bit of qubit measure_order[i]), so the counts stay correct regardless of the platform's arrangement.

QuiXamii:
> `tianyan.py` uses `dict[int, set[int]]` and `tuple[dict[int, set[int]], set[int]]` while the rest of `cloud/` (and the same file's other signatures) uses `Dict`/`Tuple` from `typing`. inconsistent within the module set.

Fixed in 0efedff6.

examples/tianyan_cloud_demo.py
QuiXamii: Done in 8ac8d39d. The hardware example now reuses the same logical Bell circuit and maps it onto a connected physical pair from device.topology() via Circuit.initial_mapping.

tensorcircuit/cloud/tianyan.py
refraction-ray: yes, please fix the typo throughout different backends, thanks,

refraction-ray:
@all-contributors please add @QuiXamii  for code, doc, test, tutorial

PR #133, opened 2026-09-13
Add NQS-tVMC example with RBM and blurred sampling

QingyunQian:
## Summary

Add a self-contained neural-network quantum state time-dependent variational
Monte Carlo (NQS-tVMC) example using TensorCircuit-NG's existing backend
infrastructure. It demonstrates the complete workflow: represent the state with
a complex RBM, sample configurations with local MCMC, estimate the quantum
geometric tensor and variational force, and evolve the parameters through TDVP.
The implementation uses `tc.backend` operations, JIT, vmap and scan.

The four-spin RBM quench is the main NQS example; a single-spin calculation
provides a simple correctness check. Published Figure 5(a,b) of
[Wan, Wiersema and Zhang, PRX 16, 031059](https://journals.aps.org/prx/abstract/10.1103/jrn5-gv19)
provides the validation benchmark. Comparing standard and blurred sampling
shows how importance reweighting restores contributions lost near wave-function
nodes within this NQS-tVMC workflow.

## Reproduction scope

Single-spin H=Y evolution and an open four-spin Heisenberg quench with a
one-hidden-unit complex RBM. No size reduction or core-library change is needed.
This contribution covers these two published panels only.

The example includes local sampling, node-safe metric and force estimators,
Schmitt regularization, RK5/Heun integration, a tiny independently regenerated
SR initial state, validation, metadata and generated gallery assets.
TensorCircuit-NG implements the evolution; NetKet is not a runtime dependency.

## Results

Figure 5(a,b): standard and blurred tVMC compared with exact dynamics

The four-spin blurred curve has maximum correlation error approximately 4.4e-6.
Ordinary tVMC first exceeds error 0.05 at figure time 0.72525, compared with
0.7275 from a matched run of the original author driver. The single-spin blurred
error is 1.8e-15. Our standard four-spin trajectory stays near zero after failure,
whereas the paper later shows oscillations; this post-failure behavior is not
matched pointwise. Raw results and measured timings are included beside the figure.

## Validation

- Black and Pylint 3.2.6 pass (10.00/10).
- Independent Pauli/TC circuit Hamiltonian, observable and evolution checks pass.
- Direct RBM derivatives agree with TC autodiff, including a cosh node.
- Blurred metric/force agree with exact amplitude inner products at nodes;
  the standard estimator's missing finite contribution is explicitly checked.
- Thirty-two MC replicas also verify S/F and the Eq. (11) covariance correction.
- Metadata/gallery generation passes.
- Repository-wide Black, Mypy and Pylint pass with the development dependencies
  installed. English and Chinese Sphinx builds pass in an offline environment;
  remote cloud notebook calls were not validated.
- Full pytest passes: 2048 passed, 53 skipped and 3 xfailed (exit code 0):
  `pytest -n auto --cov=tensorcircuit -vv -W ignore::DeprecationWarning tests/`.

## Reproduction command

```bash
python examples/reproduce_papers/2026_nqs_tvmc_blurred_sampling/main.py
python examples/reproduce_papers/2026_nqs_tvmc_blurred_sampling/validate.py
cd docs/source
python generate_gallery.py
```

Numerical runs and plotting used a Linux CPU server with JAX double precision;
per-run dependency versions, synchronized timings and peak RSS are recorded.

## Notes

The README explains the notebook's single vertical bond and physical-time/4
axis, published Table II settings for panel (a), and the source drivers' extra
factor N in force SNR. The independently regenerated SR state has ground-energy
error 0.003747 and infidelity 0.001314; both methods and exact dynamics use that
same state. Sampling, warmup and random streams differ from NetKet. Finite-sample
SNIS and regularized TDVP updates are not claimed to be unbiased.


refraction-ray:
[P1 · Performance] Replace per-step host synchronization with a backend scan

  Location: main.py:190-220

  The fixed-step four-spin Heun trajectory is JIT-compiled, but the outer Python loop calls k.numpy() at every step. This causes repeated
  device-to-host synchronization and prevents end-to-end staging. The trajectory contains 6000 steps and takes approximately 239–241
  seconds per run.

  Recommendation: implement the fixed-step trajectory with a backend scan, carrying theta, chains, key, and a trajectory buffer, and
  transfer results to NumPy only once after the scan.

  [P1 · Numerical robustness] Use a fixed step size for the RK45 benchmark

  Location: main.py:113-145 and 190-205

  Each RK45 stage performs fresh MCMC and blur sampling, so the embedded RK45 error estimate contains Monte Carlo noise in addition to
  integration truncation error. The controller can therefore reject steps because of sampling noise rather than ODE error. The author’s
  public notebook uses RK45 with adaptive=False for this benchmark, although the final paper table describes it as adaptive RK45.

  Recommendation: use the fixed step size from the author notebook and report a simple step-size convergence check. Alternatively, use
  common random numbers across RK stages.

  [P3 · Maintainability] Fix the import order

  Location: main.py:20-21 and validate.py:11-19

  The third-party tensorcircuit import appears after local physics/main imports. Under the repository .pylintrc this produces wrong-import-
  order and gives 9.96/10 rather than the claimed 10.00/10.

  Recommendation: move import tensorcircuit before the local-module imports.

examples/reproduce_papers/2026_nqs_tvmc_blurred_sampling/main.py
```diff
@@ -0,0 +1,349 @@
...
+        (gauge[:, None] * observable * gauge[None, :]),
+        hfull,
+        indices,
+        gauge,
+    )
+
+
+def exact_trajectory(problem, theta, times):
+    """Unitary reference starting from the same actual variational state."""
+    k = tc.backend
```
refraction-ray: better use K instead of k

refraction-ray:
for validate.py better make it to np.assert_allclose instead of output a separate json file with only minor error values

refraction-ray:
and we typically dont update changelog for example adding

QingyunQian:
thanks for the review. All six points are addressed

.pylintrc
```diff
@@ -4,6 +4,11 @@
 jobs=1
 
 
+[IMPORTS]
```
refraction-ray: is this change on pylintrc necessary? better keep config file intact

examples/reproduce_papers/2026_nqs_tvmc_blurred_sampling/README.md
```diff
@@ -0,0 +1,148 @@
+# NQS-tVMC with blurred sampling: Figure 5(a,b)
```
refraction-ray, line 1: this file seems to inserted \n mannually when AI generated it. Better only have \n for each natural paragraph

examples/reproduce_papers/2026_nqs_tvmc_blurred_sampling/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  - "pauli-operator"
+  - "expectation"
+  - "jit"
+  - "vmap"
+  - "scan"
+  - "autodiff"
+backend: "jax"
+hardware_requirements:
+  gpu: false
+  min_memory: "4 GB recommended; not a measured minimum"
```
refraction-ray: not a measured minimum: what does this mean?

examples/reproduce_papers/2026_nqs_tvmc_blurred_sampling/README.md
QingyunQian: Done. the manual line wrapping has been removed

.pylintrc
QingyunQian: fixed

examples/reproduce_papers/2026_nqs_tvmc_blurred_sampling/meta.yaml
QingyunQian: The wording meant that 4 GB was only a recommendation; I had not measured the minimum RAM requirement for this task. I removed that recommendation and changed min_memory to Not determined.

refraction-ray: we must have a min_memory

QingyunQian: Updated min_memory to 1 GB, estimated from the measured peak RSS of 733 MiB with approximately 30% headroom.

examples/reproduce_papers/2026_nqs_tvmc_blurred_sampling/main.py
```diff
@@ -0,0 +1,338 @@
...
+import resource
+from time import perf_counter
+from importlib.metadata import version
+
+import matplotlib
+import matplotlib.pyplot as plt
+import numpy as np
+import tensorcircuit as tc
+
+if __package__:
```
refraction-ray: no need for this? or what is the rationale of this?

examples/reproduce_papers/2026_nqs_tvmc_blurred_sampling/validate.py
```diff
@@ -0,0 +1,302 @@
...
+
+All exact sums here are validation references, never the MC evolution path.
+"""
+
+from pathlib import Path
+
+import numpy as np
+import tensorcircuit as tc
+
+if __package__:
```
refraction-ray: similar here

examples/reproduce_papers/2026_nqs_tvmc_blurred_sampling/main.py
QingyunQian: I originally added the conditional imports to support both script and module execution, but only direct script execution is needed here. It has been removed

refraction-ray:
thanks for the nice contribution

PR #134, opened 2026-09-14
Add FGSSimulator example for Chern-insulator edge dynamics

QingyunQian:
Add examples of two-dimensional real-time dynamics with `FGSSimulator` and fermionic PEPS-tVMC, based on the Chern-insulator quench in [Wu and Nys, PRX Quantum 7, 033035](https://journals.aps.org/prxquantum/abstract/10.1103/tggc-8fjx).

The FGS example retains the original 12×12, 96-particle benchmark in Figure 2(a). The PEPS example demonstrates the method behind Figure 2(b,c) on a 3×3 lattice with D=2: sampled imaginary-time VMC preparation, gauge-projected SR, and fixed-step RK4 real-time evolution. It reuses TensorCircuit-NG's contractors and backend operations and TenCirPauli's fermion algebra. FGS supplies an independent reference, and is not used to fit or initialize the PEPS.

The default PEPS calculation takes about four minutes on CPU, including preparation and evolution to t=1. Table I reports 399 seconds per step for the authors' full-size Chern calculation, implying about 5.54 days for 1,200 real-time steps alone. The small default makes the workflow practical to run and permits independent exact-state validation. It does not claim the paper's original-size accuracy or computational speedup; its gauge projector uses SVD and a dense solve rather than the paper's reduced/local QR construction.

## Validation

- Independent Hamiltonian, fermionic-sign, derivative, gauge, sampling, boundary-convergence, and time-step checks pass.
- The default PEPS run has initial energy error 6.14e-9 and phase-aligned evolution-state error 1.59e-8. Sampled density RMSE is 0.0152; sampling error is reported separately from wave-function error.
- The final scan refactor preserves the figure and numerical arrays byte-for-byte. Both validators and all 41 FGS tests pass.
- Black and Mypy pass. Pylint passes under the unchanged repository configuration, including runs from the repository root and example directory.
- The complete `check_all.sh` passed before this final import, scan, and documentation cleanup, with 2,048 tests passed, 53 skipped, and 3 xfailed. Both Sphinx builds completed with existing documentation diagnostics. The targeted checks above were rerun after cleanup.

## Run

Use an environment containing TensorCircuit-NG, JAX, NumPy, SciPy, Matplotlib, and TenCirPauli 0.5.0. From the repository root:

```bash
PYTHONPATH=. python examples/reproduce_papers/2026_chern_edge_fgs/main.py
PYTHONPATH=. python examples/reproduce_papers/2026_chern_edge_fgs/validate.py
PYTHONPATH=.:examples python examples/reproduce_papers/2026_chern_edge_fgs/run_peps.py
PYTHONPATH=.:examples python examples/reproduce_papers/2026_chern_edge_fgs/validate_peps.py
PYTHONPATH=.:examples python examples/reproduce_papers/2026_chern_edge_fgs/validate_peps.py --result-dir examples/reproduce_papers/2026_chern_edge_fgs/outputs
```

The example includes run instructions, numerical arrays, figures, metadata, and gallery assets. README.md and PEPS.md document conventions, accuracy limits, additional small-system checks, and optional larger-lattice kernel profiles.


refraction-ray:
this is subtle, as the paper's core is to demonstrate the capability of PEPS-tVMC, while you only construct an free fermion baseline which has little to do with the bulk of the paper

QingyunQian:
> this is subtle, as the paper's core is to demonstrate the capability of PEPS-tVMC, while you only construct an free fermion baseline which has little to do with the bulk of the paper

thx for pointing this out. Now I am extending the example to PEPS-tVMC for the same Chern-insulator quench, building on tc-ng existing infrastructure and retaining FGS as an exact benchmark. I think it is a good example candidate for tc-ng's time evoulution methods

refraction-ray:
some general comments on the scalability of the peps-tvmc implementation


  ### SR/Fisher-metric scalability

 The current SR implementation explicitly constructs the full parameter-space Fisher/SR metric and the dense gauge projector. This scales as (O(P^2)) in memory and becomes prohibitive for the stated 12×12, D=4 parameter count. This should be addressed before treating the implementation as scalable. Please consider minSR, a matrix-free SR/Fisher-vector product with an iterative solver such as CG, or a reduced/local gauge-coordinate construction. The gauge projection should also be applied implicitly or in a reduced space rather than by materializing a dense (P\times P) projector. Please include a memory/scaling benchmark after this change.

  ###  Contractor choice in the global/exact branch

 In the exact/global contraction branch, the implementation currently relies on TensorCircuit’s default greedy contractor. For these PEPS networks, please configure and benchmark a dedicated path finder such as OMECO (or cotengra when appropriate), and reuse/cache the resulting contraction path for repeated configurations with the same topology. The current implementation performs repeated full-network contractions for sampled and connected configurations, so leaving this branch on the default greedy path may become a significant scalability bottleneck.

PR #136, opened 2026-09-22
Fix time ordering when inverting time-dependent analog circuits

QingyunQian:
## Problem

`AnalogCircuit.inverse()` reverses the block sequence but inverts each analog
block using `-H(t)` on the original time interval. This works for a constant
Hamiltonian, but does not undo a general time-dependent evolution when
Hamiltonians at different times do not commute.

For example, evolving `|0>` under `H(t) = (1 - t) X + t Z` from 0 to 1 and
then appending `circuit.inverse()` gives a return probability of approximately
`0.92399376`, instead of 1. The forward evolution agrees with an independent
SciPy ODE solution; the error is in constructing the inverse.

## Changes

- In `tensorcircuit/analogcircuit.py`, evaluate each inverse block with
  `-H(t_start + t_end - t)`, preserving its original integration interval while
  reversing the Hamiltonian's time order.
- Capture each block's Hamiltonian and endpoint sum in its own closure.
- Update the public docstring and development guidance, and add an Unreleased
  changelog entry.
- In `tests/test_analogcircuit.py`, add coverage for local and global evolution,
  scalar durations, nonzero start times, and multiple analog blocks interleaved with digital
  gates. Add a JIT/autodiff check against a finite-difference gradient.

## Validation

- All five new round-trip regression cases fail on upstream `33361de8` before
  applying the fix.
- All five round-trip cases pass after the fix. The default JAX ODE solver's
  JIT-compiled gradient agrees with a central finite difference.
- Repository-wide Black, Mypy, and all three Pylint invocations pass.
- Full pytest: 2,171 passed, 53 skipped, 3 xfailed (exit code 0).
- English and Chinese Sphinx HTML builds pass with warnings under offline
  settings. Remote cloud notebook calls were not validated.
- `bash check_all.sh` completes with exit code 0. The Chinese Sphinx build used
  four parallel processes.


refraction-ray:
thanks for this bug fixing!
@all-contributors please add @QingyunQian for code, test

refraction-ray:
@all-contributors please add @QingyunQian for code, test

PR #141, opened 2026-09-23
Preserve density-matrix channels during copy and composition

QingyunQian:
`DMCircuit.copy()`, `append()`, and `prepend()` reconstruct circuits through `from_qir()` without enabling channel replay. Its default `allow_channel=False` silently drops recorded noise channels, so these basic operations can change the density matrix.

For example:

```python
import tensorcircuit as tc

c = tc.DMCircuit(1)
c.x(0)
c.amplitudedamping(0, gamma=0.5, p=1.0)
c.state()         # diag(0.5, 0.5)
c.copy().state()  # previously diag(0, 1)
```

This change enables channel replay at the three reconstruction sites when the destination circuit has `is_dm=True`. It preserves the default and explicit `from_qir(..., allow_channel=False)` behavior for intentionally omitting channels. The scope is channels already recorded in QIR.

Changed files:

- `tensorcircuit/abstractcircuit.py`: preserve channels during density-matrix copy/composition and document the behavior and `allow_channel` option.
- `tests/test_dmcircuit.py`: add 42 regression cases covering both density-matrix implementations, NumPy/TensorFlow/JAX, mixed inputs, channel metadata, copy independence, composition order, qubit remapping, explicit channel omission, and JIT gradients.
- `CHANGELOG.md`: record the fix.

Validation:

- Before the fix: the NumPy regression group has 8 failures and 2 passing controls.
- After the fix: related circuit/QIR tests pass (78 passed).
- Full `bash check_all.sh` completed successfully: Black, mypy, all three Pylint stages, full pytest, and English/Chinese Sphinx builds.
- Full pytest: 2279 passed, 53 skipped, 3 xfailed.
- Sphinx builds succeeded with documentation warnings (816 English, 893 Chinese); these are not warning-free builds.

Based on upstream master `689e28e8a9115e654fa31d6136759332cd3f92e9`.


refraction-ray:
very nice catches

